Issue #244, opened 2026-08-16
Conan repositories are unusable with Conan 2 clients: /v2/users/authenticate and /v2/users/check_credentials return 405


strasvet:
## Summary

A Conan **hosted** repository in Nexspence v1.33.2 implements the Conan v2 revision protocol correctly — ping, revision listing, `latest`, file listing, file upload and file download all work, and the JSON payloads match the spec exactly.

But **every upload from a Conan 2 client fails**, because the client calls `GET /v2/users/check_credentials` immediately before writing and Nexspence returns `405 Method Not Allowed` for that route (and for `/v2/users/authenticate`).

I verified that these two endpoints are the *only* gap: with them stubbed by a pass-through proxy and everything else forwarded untouched to Nexspence, a full `conan upload` → wipe cache → `conan install` round trip succeeds against real Nexspence storage, with byte-identical artifacts. Details and the exact contract below.

Re-checked on **v1.36.2** (current latest at the time of filing): the `/v2/users/*` routes still answer 405, so the report stands as written.

## Environment

| | |
|---|---|
| Nexspence | v1.33.2 (protocol testing), re-checked on v1.36.2 |
| Repository | format `conan`, type `hosted`, `allowAnonymous: true` |
| Client | Conan 2.31.2 |
| Base URL | `http://<host>:8000/repository/<conan-repo>/` |

## What already works

Verified by `PUT`-ing a probe package with `curl` and reading it back:

| Endpoint | Result |
|---|---|
| `GET /v1/ping` | `200`, `X-Conan-Server-Capabilities: revisions` |
| `GET /v2/conans/{ref}/revisions` | `404` before upload, `200` after |
| `GET /v2/conans/{ref}/latest` | `404` before upload, `200` after |
| `PUT /v2/conans/{ref}/revisions/{rrev}/files/{file}` | `201` |
| `GET /v2/conans/{ref}/revisions/{rrev}/files` | `200` |
| `GET /v2/conans/{ref}/revisions/{rrev}/files/{file}` | `200` |
| `GET .../packages/{pid}/revisions` and `/latest` | `404` before, `200` after |
| `PUT .../packages/{pid}/revisions/{prev}/files/{file}` | `201` |
| `GET .../packages/{pid}/revisions/{prev}/files` | `200` |

Response bodies are spec-correct:

```json
{"revisions":[{"revision":"aaaa…","time":"2025-01-01T00:00:00Z"}]}
[...]

PR #249, opened 2026-08-16
feat(conan): implement the v2 search and delete routes (#247)

### LINKED ISSUE #247: Conan: search and delete routes are missing (conan search / remove / retention)
Split out of #244, which covers the P0 credential endpoints (`/v2/users/authenticate`, `/v2/users/check_credentials`). These are the remaining unimplemented Conan v2 routes reported there — they answer 405 for every method because they are not registered in `internal/formats/conan/handler.go`.

| Endpoint | Breaks | Priority |
|---|---|---|
| `GET /v2/conans/search` | `conan search -r`, `conan list -r` | P1 |
| `DELETE /v2/conans/{ref}` | `conan remove -r`, retention | P1 |
| `DELETE /v2/conans/{ref}/revisions/{rrev}` | per-revision cleanup | P1 |
| `GET /v2/conans/{ref}/{version}/{user}/{channel}/search` | package-level search | P2 |

The `DELETE` routes are also what a cleanup policy would need to express retention for Conan the way the other formats do.

Reference contract: `conans/server/rest/controller/v2/` in the Conan repository.

Reported by @strasvet in #244.

strasvet:
Closes #247.

## What this adds

The remaining Conan v2 routes from the #244 report, at the scope #247 set:

- **`GET /v2/conans/search?q=<pattern>&ignorecase=<bool>`** → `{"results": [refs]}`.
  fnmatch-style pattern (`*` crosses `/` — `boost*` matches `boost/1.83.0`,
  which `path.Match` would not), case folded unless `ignorecase=False`.
  References render the way the client prints them: `name/version`, or
  `name/version@user/channel` when user/channel are real.
- **`GET /v2/conans/{ref}[/revisions/{rrev}]/search?list_only=<bool>`** →
  JSON object keyed by package ID. With `list_only=True` only the keys carry
  meaning; otherwise each value holds the RAW `conaninfo.txt` text under
  `"content"` — the client parses it itself (`load_binary_info` in
  `remote_manager.py`, "Avoid serializing conaninfo in server side"). The
  ref-level form resolves the latest recipe revision first. A recipe-only
  revision answers `{}`, which the download flow treats as "no binaries",
  not an error.
- **`DELETE /v2/conans/{ref}`** and **`DELETE /v2/conans/{ref}/revisions/{rrev}`**
  → remove the recipe (one revision or all of them), package binaries
  included, through `base.DeleteArtifact`, so blob refcounts, store usage and
  the `artifact.deleted` webhook behave like every other format; orphaned
  components are cleaned after. 200 on success, 404 when nothing was there —
  the client raises `RecipeNotFoundException` on 404, so a repeated delete
  reports honestly. These routes are also what retention needs for Conan.

The contract was taken from the Conan 2.31.2 sources directly
(`conan/internal/rest/rest_client_v2.py`, `remote_manager.py`,
`client_routes.py`, `rest_routes.py`) — response shapes, parameter literals
(`list_only=True`, `ignorecase=False`) and status expectations all match what
the client actually sends and parses, and the captured request log of a live
`conan list` / `download` / `remove` session against v1.37.0.

## Decisions worth review

- **Exhaustive listings.** `pathsUnder` now uses `Assets.ListByRepoAndPath`
  (unbounded prefix query) instead of the first page of `Assets.List`. For
  the read routes the old cap was a latent mis-listing; for the new DELETE it
  would have been silent partial deletion answered with 200 in any repository
  over 1000 assets — a few hundred packages with binaries is enough. Search
  uses the same exhaustive prefix listing.
- **v1-stored content stays out of v2 search.** Assets under `/files/...`
  (Conan 1 uploads) are not advertised: every other v2 route 404s for them,
  and a search hit that `latest`/`download` cannot resolve makes the client's
  version-range resolver pick it and abort the install it would otherwise
  complete.
- **Proxy repositories answer search from the local cache.** The generic
  proxy GET would drop the query string on the upstream fetch and cache the
  first upstream answer forever under the path key. A proxy can serve what it
  has cached, so that is what its search reports.
- **Package-granular DELETE** (`…/packages/{pid}`, `…/packages/{pid}/revisions/{prev}`)
  is not in #247's table and keeps answering 405 rather than being
  half-guessed here.
- The search pattern length is capped (256 bytes → 400): compiling it is the
  only CPU an unauthenticated caller can make this handler spend.

Access control is untouched: RBACMiddleware already maps DELETE to the
"delete" action before the format handler runs, and `RejectMutation` keeps
refusing DELETE on proxies.

## Tests

`v2_search_test.go` / `v2_delete_test.go`, same harness as the existing v2
tests: pattern shapes and case folding, `@user/channel` rendering, empty-list
answers, revision- and ref-level package search (latest-revision resolution),
raw-conaninfo `content` passthrough, `list_only`, recipe-only `{}` vs 404,
revision delete leaving siblings intact, whole-ref delete, repeated delete →
404, package-granular 405, proxy behavior for search (local) and delete
(refused), bad/oversized patterns → 400, backslash-in-class literals.

`go test ./internal/...`: conan package green, coverage 88.4% (≥80% floor);
the only failures on my machine are the known local-network-restriction cases
(`TestBlobStoreMigrationHandler_*`, `TestWebhookHandler_Test_200`), which
reproduce on a clean `main`. `golangci-lint run` (v2.12.2): 0 issues.

Reported and verified with a live Conan 2.31.2 client against our v1.37.0
deployment — `conan remote login` / `upload` / `install` work there today, and
this PR is what `conan list -r` / `download` / `remove` were still missing.

### DIFF
--- a/internal/formats/conan/handler.go
+++ b/internal/formats/conan/handler.go
@@ -65,6 +65,19 @@ func (h *Handler) ServeHTTP(c *gin.Context) {
 			h.serveUsers(c, p)
 			return
 		}
+		// Search too (#247): the generic proxy GET below would drop the query
+		// string on the upstream fetch and cache the first answer forever, so
+		// forwarding it can only mislead. A proxy can serve what it has
+		// cached, and that is what its search reports.
+		if c.Request.Method == http.MethodGet && p == "/v2/conans/search" {
+			h.v2SearchRecipes(c, repoName)
+			return
+		}
+		if c.Request.Method == http.MethodGet &&
+			strings.HasPrefix(p, "/v2/conans/") && strings.HasSuffix(p, "/search") {
+			h.v2SearchPackages(c, repoName, p)
+			return
+		}
 		// Conan revision files are content-addressed (immutable); the "/latest"
 		// endpoints resolve to a moving recipe/package revision, so revalidate those.
 		var maxAge time.Duration
--- a/internal/formats/conan/v2.go
+++ b/internal/formats/conan/v2.go
@@ -7,6 +7,9 @@ package conan
 //	GET  /v2/conans/:name/:ver/:user/:channel/revisions/:rrev/files       → list recipe files
 //	GET  /v2/conans/:name/:ver/:user/:channel/latest                      → latest recipe revision
 //	GET  /v2/conans/:name/:ver/:user/:channel/revisions                   → list recipe revisions
+//	GET  /v2/conans/search?q=<pattern>                                    → recipe search (v2_search.go, #247)
+//	GET  /v2/conans/:name/:ver/:user/:channel[/revisions/:rrev]/search    → package search (v2_search.go, #247)
+//	DELETE /v2/conans/:name/:ver/:user/:channel[/revisions/:rrev]         → delete recipe / revision (v2_delete.go, #247)
 //
 // ...and the same shapes under /revisions/:rrev/packages/:pkgid/... for
 // package binaries. Files are stored verbatim under their full
@@ -44,6 +47,18 @@ func (h *Handler) serveV2(c *gin.Context, repoName, p string) {
 	case m == http.MethodGet && strings.HasSuffix(p, "/revisions"):
 		h.v2ListRevisions(c, repoName, p)
 
+	// Repository-wide recipe search: /v2/conans/search?q=<pattern> (#247)
+	case m == http.MethodGet && p == "/v2/conans/search":
+		h.v2SearchRecipes(c, repoName)
+
+	// Binary package search, ref- or revision-level: .../search (#247)
+	case m == http.MethodGet && strings.HasSuffix(p, "/search"):
+		h.v2SearchPackages(c, repoName, p)
+
+	// Recipe deletion, whole reference or one revision (#247)
+	case m == http.MethodDelete:
+		h.v2Delete(c, repoName, p)
+
 	default:
 		c.Status(http.StatusMethodNotAllowed)
 	}
@@ -52,17 +67,20 @@ func (h *Handler) serveV2(c *gin.Context, repoName, p string) {
 // pathsUnder returns stored asset paths (relative to prefix) and their
 // timestamps for every asset under the given path prefix. The second return
 // is false when listing failed and an error response was already written.
+//
+// The listing is exhaustive: ListByRepoAndPath is an unbounded prefix query,
+// unlike the paged Assets.List, whose first page is all a handler would see.
+// That distinction is live for DELETE (#247) — deleting from a truncated
+// listing would remove part of a revision and still answer 200.
 func (h *Handler) pathsUnder(c *gin.Context, repoName, prefix string) (map[string]time.Time, bool) {
-	page, err := h.deps.Assets.List(c.Request.Context(), repoName, 1000, 0)
+	assets, err := h.deps.Assets.ListByRepoAndPath(c.Request.Context(), repoName, prefix)
 	if err != nil {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 		return nil, false
 	}
 	out := map[string]time.Time{}
-	for _, a := range page.Items {
-		if strings.HasPrefix(a.Path, prefix) {
-			out[strings.TrimPrefix(a.Path, prefix)] = a.LastModified
-		}
+	for _, a := range assets {
+		out[strings.TrimPrefix(a.Path, prefix)] = a.LastModified
 	}
 	return out, true
 }
--- a/internal/formats/conan/v2_delete.go
+++ b/internal/formats/conan/v2_delete.go
@@ -24,7 +24,10 @@ import (
 //
 // Package-granular deletion (…/packages/:pkgID, …/packages/:pkgID/revisions/:prev)
 // is not part of #247 and keeps answering 405 rather than being half-guessed
-// here.
+// here. Content a Conan 1 client stored under /files/… is likewise out of
+// scope, the same way it is for search: these routes only ever see the
+// /v2/conans/… tree, so a retention job built on them expires v2 revisions
+// and leaves v1 files where they are.
 //
 // Write access is the caller's problem to have: RBACMiddleware maps the DELETE
 // method to the "delete" action before the format handler runs, and proxy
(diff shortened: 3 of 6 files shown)


skensell201:
Thanks for this — and for the report that led to it. The contract work is visible: the response shapes, the `list_only=True` / `ignorecase=False` literals and the raw-`conaninfo` passthrough all match what the 2.x client actually sends and parses, and the reasoning in the PR body about *why* v1-stored content stays out of v2 search is the right call for the same reason the version-range resolver would otherwise abort an install.

I reviewed it against a server built from the branch, with a real Postgres behind it rather than the in-memory asset repository. Notes below, one blocking-ish and the rest for your judgement.

## The exhaustive-listing switch needs #250 first

Moving `pathsUnder` to `Assets.ListByRepoAndPath` is the right instinct — the first-page cap was a latent mis-listing, exactly as you say. But that query built its pattern as `a.path LIKE $prefix || '%'` with no `ESCAPE` clause, so LIKE metacharacters in the prefix widened it. That is not a hypothetical here: `_` is the Conan placeholder for an absent user/channel, so *every* reference without a user/channel goes through it. On the branch:

```
GET /v2/conans/my_lib/1.0/_/_/revisions/r1/files
{"files":{"/v2/conans/myXlib/1.0/a/b/revisions/r1/files/conanfile.py":{},
          "conanfile.py":{},"conanmanifest.txt":{}}}
```

An unrelated recipe's absolute path in a plain file listing — the client reads that and asks for a file that does not exist. With a percent-encoded segment (`%25`) it enumerates every path in the repository.

Deletion turns out to be safe: `pathsUnder` trims the prefix off, a wildcard-matched path does not carry that prefix literally, so the reconstructed path misses and `DeleteArtifact` is idempotent. I checked — `DELETE /v2/conans/%25/%25/%25/%25` answers 200 and deletes nothing, and `DELETE …/my_lib/1.0/_/_` leaves `myXlib/1.0@a/b` intact. So it is a read-side problem, not a destructive one. It does make DELETE answer 200 for a reference that does not exist, though.

None of this is yours to fix — the hole is in the repository layer and predates the PR (the Docker browse handlers pass an image name straight into it). #250 escapes the prefix and adds `ESCAPE '\'`, with an integration test. Once that lands, rebase and this goes away. Worth knowing why your tests could not have caught it: `testutil.AssetRepo` matches with `strings.HasPrefix`, which is the semantics the SQL was *supposed* to have.

## Smaller things

- **Partial deletion behind a 500.** `v2Delete` returns on the first `DeleteArtifact` error, so a failure on file three of ten leaves the revision half-gone while `revisions`/`latest` still advertise it. Since `DeleteArtifact` is idempotent, attempting all of them and reporting the first error afterwards makes a client retry converge instead of chasing a moving target.

- **v1-stored content is not deleted either.** The prefix is always `/v2/conans/…`, but a Conan 1 upload lands under `/files/{n}/{v}/{u}/{c}/…`. For a reference that only exists there, `DELETE /v2/conans/{ref}` answers 404 and the bytes stay. Given the PR body offers these routes as what retention needs, either sweep the v1 prefix too or say in the doc comment that v1 content is out of scope for deletion the way it is for search.

- **Repository-wide search materializes every asset row.** `ListByRepoAndPath(repo, "/v2/conans/")` returns full `domain.Asset` structs across two joins, and the handler uses only `Path`. Version-range resolution issues a search on every `conan install`. `Assets.ListRawAssetPaths` is the path-only projection.

- **`[^abc]` negates; fnmatch treats it as a literal `^`.** `q=[^m]*` returns `secretlib/9.9@team/stable` on the branch; Python's `fnmatch.translate` escapes a leading `^` and would match the `m…` references instead. `[!…]` is handled correctly. Tiny, but it is the one place the translation diverges from the module conan_server uses.

- **`packageInfo` returns a bare `{}` on a read failure**, which is byte-identical to the `list_only=True` shape. `RemoteManager.search_packages` unwraps `data["content"]` for every entry of a non-`list_only` answer, so a package revision missing `conaninfo.txt` may become a `KeyError` client-side rather than a package reported without settings. Worth a check against the 2.31.2 sources you were reading — `{"content": ""}` or omitting the ID would degrade more predictably.

- **Proxy search answering from the local cache** is a real behaviour change: `conan list -r <proxy>` on a proxy that has not cached anything now reports nothing rather than the upstream's contents. Your reasoning for not forwarding is right — the query string would be dropped and the answer cached forever under the path key — so I am not asking you to change it, but it should land in the release notes as such.

## Checked and fine

The DELETE shape validation leaves package-granular deletes at 405, including the `…/packages/{pid}/revisions/{prev}` form. `RejectMutation` still runs before the new proxy search interception. The `serveV2` case ordering does not shadow `/files/`, `/latest` or `/revisions` — a file literally named `search` still downloads (on a hosted repository; on a proxy the new interception does win, if you want them consistent). `fnmatchRegexp`'s anchoring, backslash-in-class escaping and unterminated-`[` fallback are correct, and RE2 makes the 256-byte cap enough — a pattern of 128 `*`s costs nothing.

skensell201:
Sorry — my review above is a lot of prose and no verdict. Here is the short version.

## What happened, and why it is not on you

Your switch from the paged `Assets.List` to `Assets.ListByRepoAndPath` was the right call, for exactly the reason you gave. It just happened to be the first caller that fed a *caller-controlled* prefix into that query. `ListByRepoAndPath` built its filter as `a.path LIKE $prefix || '%'` with no `ESCAPE` clause, so LIKE metacharacters in the prefix widened the match — and `_`, the Conan placeholder for an absent user/channel, is a LIKE metacharacter. Every reference without a user/channel went through it.

Two reasons nobody caught this earlier:

- The existing callers were the Docker browse handlers, where a wildcarded listing looks like a browse quirk rather than a bug. It was already broken there; nothing pointed at it.
- `testutil.AssetRepo` matches prefixes with `strings.HasPrefix` — which is the semantics the SQL was *supposed* to have. So the test double was right and the database was wrong, and no unit test on either side could see the difference. I only hit it because I ran your branch against a real Postgres.

So: a pre-existing hole in the repository layer, surfaced by a correct change of yours. I fixed it separately in #250 (merged) rather than asking you to carry an unrelated fix through your PR.

## What this needs from you

1. **Rebase on `main`.** That is the whole fix for the listing problem — none of your code changes.
2. **`v2Delete`: attempt every asset, report the first error at the end**, instead of returning on the first failure. `DeleteArtifact` is idempotent, so a client retry then converges instead of chasing a half-deleted revision that `revisions`/`latest` still advertise.
3. **One line in the doc comment** saying v1-stored content (`/files/…`) is out of scope for deletion, the way you already say it is for search. Your PR body offers these routes as what retention needs, and I want the gap written down where the next person will read it.

Everything else in my review is your call — the `[^abc]` translation, the `packageInfo` shape on a failed read, moving the repository-wide search to `ListRawAssetPaths`. Fine to land without any of them, fine to fold them in if you would rather.

Workflow runs are approved now, so CI runs on push from here on. Once it is green I will squash it in with your authorship intact. The release note for the proxy-search behaviour change is on me.

Thanks for the report in #244 and for picking this up — the contract research behind it is the part that would have taken me longest.

strasvet:
Rebased on main (picking up #250 — confirmed the `ESCAPE '\'` is in the query now) and pushed the requested changes:

- **v2Delete attempts every asset and reports the first error at the end** — a retry after a mid-list failure now converges on 404 instead of chasing a half-deleted revision.
- **Doc comment states the v1 scope gap for deletion** the same way search does: these routes only see the `/v2/conans/…` tree, so retention built on them expires v2 revisions and leaves v1 files where they are.
- From the optional list, took the class translation fix: a leading `^` is now a literal caret (`[^b]` matches `b`), matching Python's fnmatch; backslashes inside classes are escaped before the prefix handling so the two rules compose.

Left `packageInfo`'s failed-read shape and the `ListRawAssetPaths` move as they were.

And thanks for the #250 analysis — the mock-vs-Postgres note especially; good to know exactly why the unit tests were structurally blind to it.

PR #388, opened 2026-09-04
fix: enforce RBAC on browse-docker delete endpoints

jorgemillan:
DELETE /api/v1/browse/repositories/:name/{path,docker-tag,docker-image} called base.DeleteArtifact straight from the handler with no RBAC check, unlike the standard protocol's DELETE /repository/:repoName/*path (which goes through RBACMiddleware -> CanAccessRepo(...,"delete")). Any authenticated user, regardless of privilege, could delete artifacts in any repository just by knowing its name.

Adds an authorizeDelete helper mirroring RBACMiddleware's own check (401 unauthenticated / 403 forbidden) and calls it at the top of all three handlers. Also adds /api/v1/browse/ to isAuditablePath so a legitimate delete through this API leaves an audit record.

Verified against a live stack (real Postgres, real MinIO): a privilege-less authenticated user gets 403 and the artifact survives; an anonymous caller gets 401; admin gets 204 and the artifact is gone.

### DIFF
--- a/internal/api/audit_middleware.go
+++ b/internal/api/audit_middleware.go
@@ -84,6 +84,7 @@ func isAuditablePath(path string) bool {
 		"/service/rest/v1/blobstores",
 		"/service/rest/v1/cleanup-policies",
 		"/api/v1/webhooks",
+		"/api/v1/browse/",
 		"/api/v1/login",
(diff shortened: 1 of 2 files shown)


skensell201:

PR #389, opened 2026-09-04
fix: close re-upload races in DeleteArtifact/cleanup and blob-store m…

### LINKED ISSUE #377: `AssetRepo.Create`'s upsert doesn't keep `blob_store_id`/`last_modified` correct — breaks re-upload safety checks and cross-store migration
Three related gaps in `AssetRepo.Create`'s `ON CONFLICT` upsert and its `RETURNING` clause, found together while tracing how a concurrent re-upload interacts with deletion and blob-store migration:

1. **The re-upload guard never actually detects a re-upload.** Both `base.DeleteArtifact` and `CleanupService.runPolicy` protect against a client re-uploading a path while a delete for the *old* content is in flight, by re-reading the asset under the blob-key advisory lock and comparing it to the snapshot taken before the lock. The comparison uses `BlobKey` — but `BlobKey` is `sha256(repoName + ":" + filePath)`, entirely content-independent: a re-upload to the same path always produces the *same* `BlobKey`. The guard is a no-op for the exact race it was written for (`#325`'s "downloaded or re-uploaded" case only ever covered the download half in practice).
2. **`blob_store_id` is never updated on conflict.** During a blob-store migration, the repository's `blob_store_id` stays pointed at the source store for the whole run, so an ordinary write during the migration always physically lands on source — but `Create`'s `ON CONFLICT ... DO UPDATE SET` omits `blob_store_id`, so a re-upload's own correctly-resolved store pointer gets silently discarded by whichever `Create`/`UpdateBlobStoreForBlobKey` call happens to run last.
3. **`RETURNING` never includes `last_modified`**, even though the `INSERT`/`ON CONFLICT` always sets it. Any caller that constructs an asset, calls `Create`, and reads `LastModified` from the same in-memory struct afterward (instead of re-fetching) gets a zero-value timestamp instead of the real one — a correctness trap for any future fix that relies on `last_modified` (including a fix for gap #1 above).

**Where**
- `internal/repository/postgres/asset_repo.go` (`Create`) — `ON CONFLICT` omits `blob_store_id`; `RETURNING id, created_at` omits `last_modified`
- `internal/formats/base/store.go` (`DeleteArtifact`) — lock-scoped re-read compares only `cur.ID`/`cur.BlobKey`
- `internal/repository/postgres/asset_repo.go` (`DeleteIfUnchanged`) — `WHERE` clause has no `last_modified` term
- `internal/service/cleanup_service.go` — calls `DeleteIfUnchanged` with the stale snapshot
- `internal/service/blob_store_migration_service.go` — per-row `UpdateBlobStoreForBlobKey`/`dropSourceCopy`, unlocked against a concurrent write to the same key

**Impact**

Real data loss and silent corruption: a cleanup scan or manual delete racing a re-upload to the same path can delete the newly-uploaded content; a re-upload racing a blob-store migration can end up with the DB row pointing at one store while the real bytes sit on the other — both silent, no error surfaced anywhere.

**Suggested fix**

- Add `blob_store_id = EXCLUDED.blob_store_id` and `last_modified` to `RETURNING` in `Create`.
- Add `last_modified` to `DeleteIfUnchanged`'s `WHERE` clause and to `DeleteArtifact`'s lock-scoped re-read guard.
- Serialize the migration's per-row flip+drop against other same-key actors via the existing `WithBlobKeyLock` primitive (already used by `DeleteArtifact`, `CleanupService`, `oci.mountBlob`), re-verifying the source object's size right before the flip. Note: closing this fully also needs the write path (`RegisterStoredBlob`/`StoreArtifact`) to take the same lock — but `oci.mountBlob` already nests a `RegisterStoredBlob` call *inside* its own `WithBlobKeyLock` for the same key, so an unconditional lock in the write path would self-deadlock that call site; that nesting needs resolving first.

jorgemillan:
Closes #377

Three compounding bugs around AssetRepo.Create's upsert:

1. DeleteArtifact's and CleanupService's re-upload guards compared BlobKey, but BlobKey is sha256(repoName+":"+filePath) - content independent, so a re-upload to the same path always produces the same BlobKey. The guard was a no-op for the exact race it exists for. Fixed by adding last_modified (which Create's upsert always bumps) to both DeleteIfUnchanged's WHERE clause and DeleteArtifact's lock-scoped re-read guard.

2. Create's RETURNING clause never returned last_modified, so a freshly created asset's Go struct carried a zero-value timestamp instead of the real one - fixed by adding it to RETURNING/Scan.

3. Blob-store migration's per-row flip+drop (UpdateBlobStoreForBlobKey + dropSourceCopy) ran unlocked and Create's ON CONFLICT never updated blob_store_id, so a client re-uploading a path mid-migration could end up with the DB row pointing at one store while the new bytes landed on the other. Fixed by adding blob_store_id = EXCLUDED.blob_store_id to the upsert, and running the flip+drop inside the existing WithBlobKeyLock primitive with the source object's size re-verified right before the flip.

Verified: a mocked-race unit test for DeleteArtifact (injects a same-blob-key re-upload between its two reads), a real-Postgres integration test for DeleteIfUnchanged reproducing the actual production sequence (same blob_key, only last_modified differs - added alongside the existing, narrower blob_key-change test), and a live compose smoke test confirming the normal (non-racing) delete path is unaffected. go build/go vet/go test ./... (unit + -tags=integration) green.

### DIFF
--- a/internal/formats/base/store.go
+++ b/internal/formats/base/store.go
@@ -478,7 +478,11 @@ func DeleteArtifact(ctx context.Context, d formats.Deps, repoName, filePath stri
 		if gerr != nil {
 			return gerr
 		}
-		if cur == nil || cur.ID != asset.ID || cur.BlobKey != asset.BlobKey {
+		// BlobKey alone doesn't catch a same-path re-upload: it's
+		// sha256(repoName+":"+filePath), content-independent, so a re-upload to
+		// this exact path always keeps the same BlobKey. last_modified is what
+		// actually changes.
+		if cur == nil || cur.ID != asset.ID || cur.BlobKey != asset.BlobKey || !cur.LastModified.Equal(asset.LastModified) {
 			return nil
 		}
 		// One blob can carry several assets: an OCI manifest push registers the tag
--- a/internal/repository/interfaces.go
+++ b/internal/repository/interfaces.go
@@ -65,10 +65,10 @@ type AssetRepo interface {
 	Create(ctx context.Context, a *domain.Asset) error
 	Delete(ctx context.Context, id string) error
 	// DeleteIfUnchanged deletes the row only if it is still exactly what an
-	// earlier scan read — same blob key, same last_downloaded. It reports
-	// whether the row was deleted; a mismatch (the asset was downloaded or
-	// re-uploaded since the scan) leaves the row alone.
-	DeleteIfUnchanged(ctx context.Context, id, blobKey string, lastDownloaded *time.Time) (bool, error)
+	// earlier scan read — same blob key, same last_downloaded, same
+	// last_modified. It reports whether the row was deleted; a mismatch (the
+	// asset was downloaded or re-uploaded since the scan) leaves the row alone.
+	DeleteIfUnchanged(ctx context.Context, id, blobKey string, lastDownloaded *time.Time, lastModified time.Time) (bool, error)
 	// WithBlobKeyLock serializes every caller that wants to read-then-act on a
 	// given blob key (count references, physically delete, register a new
 	// reference). Any other WithBlobKeyLock caller for the same key — same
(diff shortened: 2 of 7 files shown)


skensell201:

PR #390, opened 2026-09-04
fix: recover panics in background goroutines and cron jobs

### LINKED ISSUE #378: No goroutine in the codebase recovers from a panic
`grep -rn "recover()"` across the whole repository (excluding tests) returns zero matches. Go terminates the entire process on any unrecovered goroutine panic — not just the goroutine that panicked — so a single bad input reaching any background job (a malformed response from a source Nexus instance being migrated, an unexpected shape from Trivy/OSV.dev, a nil dereference in a cron job body) takes the whole server down, every in-flight request included, not just the misbehaving job. `robfig/cron/v3` does not recover job panics by default either — the `cron.Recover` chain wrapper has to be added explicitly, and none of this codebase's three plain `cron.New()` calls do.

**Where**

Cron schedulers without `cron.Recover`: `internal/service/cleanup_service.go`, `gc_service.go`, `replication_service.go` (a fourth, `scan_service.go`, already has `SkipIfStillRunning` but not `Recover`). Detached goroutines with no panic guard: every `go ...`/`go func(){...}()` launch site across `internal/api/router.go` (cron scheduler starts, download-counter loop, migration `ResumeAll` calls), `nexus_migration_service.go` and `blob_store_migration_service.go` (per-job run goroutines), `internal/api/handlers/cleanup.go` and `replication.go` (manual-trigger and reload goroutines), and several request-adjacent ones: `internal/api/audit_middleware.go` (per-request audit write), `internal/api/ratelimit_middleware.go` (periodic bucket eviction), `internal/api/handlers/system.go` and `health.go` (parallel health-check probes — these `wg.Wait()` on the goroutine, which does **not** protect the process: a panic in a child goroutine crashes the whole program regardless of whether the parent is waiting on it), `internal/service/webhook_service.go` (webhook dispatch), `internal/metrics/ring_buffer.go` (metrics sampler), `cmd/server/main.go` (audit partition rotator).

**Impact**

A single malformed input to any background subsystem is a full denial-of-service for the whole instance, far beyond that job's own blast radius.

**Suggested fix**

A small `safego.Go(log, name, fn)` helper that runs `fn` in a goroutine behind a `defer recover()` that logs instead of crashing, used at every detached-goroutine call site; `cron.WithChain(cron.Recover(cron.DefaultLogger))` on the three bare `cron.New()` calls. For a persistent loop goroutine that must keep running after one bad iteration (not just survive the process), the recover has to sit *inside* the loop body, not around the whole goroutine.

jorgemillan:
Closes #378

grep -rn "recover()" across the whole repository (excluding tests) returned zero matches. Go terminates the entire process on any unrecovered goroutine panic - not just the goroutine that panicked - so a single bad input reaching any background job (a malformed response from a source Nexus instance being migrated, an unexpected shape from Trivy/OSV.dev, a nil dereference in a cron job body) took the whole server down, every in-flight request included. robfig/cron/v3 does not recover job panics by default either - the cron.Recover chain wrapper has to be added explicitly, and none of the three plain cron.New() calls had it.

Adds internal/safego: Go(log, name, fn) for one-shot detached goroutines (a background job, a fire-and-forget write), Recover(log, name) as a defer for a persistent loop iteration (a cron job body, a ticker-driven sampler) so one bad iteration doesn't take the next ones down with it. Applied at every detached-goroutine and parallel-checks call site: cron scheduler starts, the download-counter and metrics- sampler loops, migration ResumeAll/run, cleanup/replication manual- trigger and reload goroutines, the audit-write and webhook-dispatch goroutines, the audit partition rotator, the rate-limit bucket-eviction loop, and the parallel service/readiness health checks (whose wg.Wait() does not itself protect the process from a child panic). Left out deliberately: the two io.Pipe copy goroutines in store.go/repoproxy.go (tight io.Copy over hash writers, not complex/adversarial-data parsing) and main.go's HTTP listener goroutine (already funnels a real listen failure to os.Exit(1) by design).

cron.WithChain(cron.Recover(cron.DefaultLogger)) added to the three bare cron.New() calls (cleanup, gc, replication schedulers) and merged into scan_service.go's existing SkipIfStillRunning chain, Recover outermost so a job panic can't also skip SkipIfStillRunning's own cleanup.

A few constructors gained a logger parameter/WithLogger option where none existed (RateLimitMiddleware, AuditMiddleware, ReadinessHandler, SystemHandler, ReplicationHandler, BlobStoreMigrationService, WebhookService, metrics.StartSampler) - existing tests updated to pass nil, which safego handles safely (skips logging, still recovers).

Verified: internal/safego/safego_test.go (a panicking goroutine is swallowed - the test itself would otherwise crash the test binary; a nil logger doesn't itself panic; a loop survives one panicking iteration). Full suite green: go build/go vet/go test ./... (unit + -tags=integration against a real dockertest Postgres). Live compose smoke test (real Postgres/MinIO/Trivy): server boots clean, /healthz and /readyz OK, logged in and browsed Repositories and System Admin -> Service Connections (the parallel-check path this fix touches) in Chrome with zero console errors.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/nexspence-oss/nexspence/internal/metrics"
 	"github.com/nexspence-oss/nexspence/internal/repository"
 	"github.com/nexspence-oss/nexspence/internal/repository/postgres"
+	"github.com/nexspence-oss/nexspence/internal/safego"
 	"github.com/nexspence-oss/nexspence/internal/storage"
 	"github.com/nexspence-oss/nexspence/internal/tracing"
 )
@@ -159,7 +160,7 @@ func cmdServe() *cobra.Command {
 			// current month's partition exists before we accept traffic.
 			rotator := audit.NewRotator(audit.NewPgPartitionStore(pool), cfg.Audit, log)
 			rotator.RunOnce(cmd.Context())
-			go rotator.Run(cmd.Context())
+			safego.Go(log, "audit-rotator", func() { rotator.Run(cmd.Context()) })
 			log.Info("audit rotator started",
 				"retention_days", cfg.Audit.RetentionDays,
 				"soft_cap", cfg.Audit.SoftCap,
@@ -195,7 +196,7 @@ func cmdServe() *cobra.Command {
 			// Start background metrics sampler — stops on context cancellation.
 			samplerCtx, cancelSampler := context.WithCancel(cmd.Context())
 			defer cancelSampler()
-			metrics.StartSampler(samplerCtx, pool)
+			metrics.StartSampler(samplerCtx, log, pool)
 
 			// Lifetime of everything the router runs in the background. Kept
 			// separate from the HTTP server's shutdown context so the two can be
--- a/internal/api/audit_middleware.go
+++ b/internal/api/audit_middleware.go
@@ -7,12 +7,14 @@ import (
 	"github.com/gin-gonic/gin"
 
 	"github.com/nexspence-oss/nexspence/internal/domain"
+	"github.com/nexspence-oss/nexspence/internal/logger"
 	"github.com/nexspence-oss/nexspence/internal/repository"
+	"github.com/nexspence-oss/nexspence/internal/safego"
 )
 
 // AuditMiddleware writes an audit event after each mutating request completes.
 // It only records PUT/POST/DELETE/PATCH on key management paths.
-func AuditMiddleware(auditRepo repository.AuditRepo) gin.HandlerFunc {
+func AuditMiddleware(log logger.Logger, auditRepo repository.AuditRepo) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		c.Next() // run handler first
 
@@ -70,7 +72,7 @@ func AuditMiddleware(auditRepo repository.AuditRepo) gin.HandlerFunc {
 			Context:    ctxData,
 			Result:     result,
 		}
-		go func() { _ = auditRepo.Write(context.Background(), e) }()
+		safego.Go(log, "audit-write", func() { _ = auditRepo.Write(context.Background(), e) })
 	}
 }
 
--- a/internal/api/audit_middleware_test.go
+++ b/internal/api/audit_middleware_test.go
@@ -19,7 +19,7 @@ func init() { gin.SetMode(gin.TestMode) }
 
 func buildAuditRouter(auditRepo *testutil.AuditRepo) *gin.Engine {
 	r := gin.New()
-	r.Use(api.AuditMiddleware(auditRepo))
+	r.Use(api.AuditMiddleware(nil, auditRepo))
 
 	r.POST("/service/rest/v1/repositories", func(c *gin.Context) {
 		c.JSON(http.StatusCreated, gin.H{"ok": true})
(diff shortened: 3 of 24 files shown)


skensell201:
Thanks for this one — the audit is right, and `internal/safego` is a good shape for it (the `Go` / `Recover` split, and the note about `cron.Recover` having to sit outermost in `scan_service.go`'s chain, are both correct).

Two things I'd like fixed before merging.

**1. `readinessHandler`: a panicking probe now reports *ready*.**

In `internal/api/handlers/health.go`, `checks[name]` is only written *after* `p.Ping` returns. `safego.Recover` swallows the panic, so the check records nothing at all and `failed` stays `false` — the probe drops out of the response and `/readyz` answers `200 {"status":"ok"}` for a dependency that just blew up. Before this PR the process crashed, which is worse in isolation but at least visible to an orchestrator; now a load balancer keeps routing traffic to a node whose Postgres check is panicking on every probe.

Reproduced on your branch:

```go
type panickingPinger struct{}

func (panickingPinger) Ping(context.Context) error { panic("probe blew up") }

func TestReadiness_PanickingProbeReportsNotReady(t *testing.T) {
	h := readinessHandler(nil, panickingPinger{}, nil, time.Minute)
	w := serveReady(h)
	if w.Code != http.StatusServiceUnavailable {
		t.Fatalf("want 503 got %d body=%s", w.Code, w.Body.String())
	}
}
```

```
health_panic_probe_test.go:21: want 503 got 200 body={"status":"ok"}
```

Recording the result from a `defer` that starts pessimistic fixes it — the deferred recorder is registered *before* `safego.Recover`, so LIFO runs `Recover` first and the recorder still sees `status == "error"`:

```go
check := func(name string, p pinger) {
	defer wg.Done()
	// The result is recorded from a defer and starts as "error":
	// safego.Recover below swallows a panic in Ping, and a check that
	// recorded nothing would drop out of the response entirely, so
	// /readyz would answer 200 "ready" for a dependency that just blew
	// up. A successful ping overwrites it right before returning.
	status := "error"
	defer func() {
		cmu.Lock()
		checks[name] = status
		if status != "ok" {
			failed = true
		}
		cmu.Unlock()
	}()
	defer safego.Recover(log, "readiness-check-"+name)()
	pctx, cancel := context.WithTimeout(ctx, 3*time.Second)
	defer cancel()
	if err := p.Ping(pctx); err == nil {
		status = "ok"
	}
}
```

Worth adding the test above alongside it.

**2. A recovered panic in a detached migration leaves the record stuck in `running`.**

`safego.Go(s.log, "blob-store-migration-run", ...)` and `safego.Go(s.log, "nexus-migration-run", ...)` now swallow a panic that previously took the process down, but neither path marks the migration failed on the way out — the row keeps saying `running` forever, and `cancels` keeps its entry. Same shape as the `flipErr` branch just below in `runMigration`: a `defer` that calls `FinishMigration(bgCtx, m.ID, "failed", &msg)` when it recovers would close it. Not a blocker for the panic-safety goal itself, but it turns a loud failure into a silently wedged job.

Everything else looks right to me: the `defer wg.Done()` ordering, the nil-logger handling, and leaving out the two `io.Pipe` copy goroutines and `main.go`'s listener are all the calls I'd have made. Verified locally alongside the other seven PRs in this batch: `go build` / `go vet` clean, full suite green.

jorgemillan:
Fixed — status now defaults to "error" and is recorded from a defer registered after cancel(), so it runs before safego.Recover's defer (LIFO) and captures the pessimistic default even when Ping panics. Added TestReadiness_PanickingDependencyReportsFailure, confirmed it fails on the old code (200 "ok") and passes with the fix (503, "error").

Issue #291, opened 2026-08-20
`ComponentHandler.Get` has no RBAC/content-selector filter, unlike `List`/`Search` (IDOR)


jorgemillan:
## Summary

`List` (`internal/api/handlers/components.go:54-125`) and `Search`/`SearchAssets` all call `h.rbacSvc.FilterComponents`/`FilterAssets` before returning results (`List` does so at `:110`, using `allowAnonMap` for each repository's `AllowAnonymous` flag). `Get` (`:157-176`) does none of this — it loads the component by ID and returns it in full to any authenticated caller, with no check of the repository's `AllowAnonymous` setting or the caller's privileges/content-selectors over that repository.

## Where

`internal/api/handlers/components.go:54-125` (`List`, with the RBAC filter at `:110`) contrasted with `:157-176` (`Get`, no filter at all)

## Reproduction (confirmed live)

A non-anonymous repository had a component uploaded to it. A normal user with no role or privilege over that repository was used to make two requests: `GET /service/rest/v1/components?repository=<repo>` returned `{"items":[]}` (correctly filtered). `GET /service/rest/v1/components/<id>` (same component, same token) returned `HTTP 200` with the full component, including its assets, `downloadUrl`, and — critically — the persisted vulnerability scan result under `extra.scan_result`, complete with the real CVEs.

## Impact

Any authenticated user who knows (or guesses/enumerates) a component UUID can read its full metadata and its security report, with no privilege whatsoever over the repository it belongs to. The actual blob download under `/repository/...` remains protected, but the metadata and scan report do not.

## Suggested fix

In `Get`, after loading the component, apply the same filter `List`/`Search` already apply (the repository's `AllowAnonymous` flag plus `FilterComponents`/a privilege check) and return 404/403 if the filter would exclude it.

**Verified:** implemented the suggested fix exactly — after loading the component, `Get` runs it through `h.rbacSvc.FilterComponents` (the same call `List` already makes) and returns `404` if the filter excludes it. `go build`/`go vet`/`go test ./...` green. Live-verified against a rebuilt `docker compose` instance: created a non-anonymous raw repo, uploaded a real asset as admin, then as a freshly created privilege-less user confirmed `GET .../components?repository=...` still correctly returns `{"items":[]}` while `GET .../components/<id>` now returns `404` — where before the fix it returned the full component (assets, `downloadUrl`, scan data) with no privilege check at all.

skensell201:
[...]

PR #391, opened 2026-09-04
fix: RBAC missing on four sibling read endpoints

### LINKED ISSUE #379: RBAC missing on several read endpoints (scan results, assets/:id, repository Get, promotion rules)
Four endpoints that each have a sibling doing the right thing, but skip the RBAC filter/check that sibling has:

- **`GET /components/:id/scan`** — unlike `ComponentHandler.Get`, which filters by RBAC and cites `#291` ("a direct GET by id must answer the same question... or knowing a UUID becomes a key to metadata and scan results the caller has no privilege over"), `ScanHandler.GetScanResult` never checks repository visibility before returning a component's full Trivy scan result.
- **`GET /assets/:id`** — an inline handler in `router.go` calls `assetRepo.Get` directly and returns the full asset (path, checksums, `DownloadURL`) with no RBAC check, unlike `SearchAssets`/`FilterAssets`.
- **`RepositoryHandler.Get`** — `RepositoryHandler.List` and `ComponentHandler.GetQuota` both filter by RBAC (`GetQuota` citing `#295`); the sibling `Get`-by-name endpoint doesn't, letting any authenticated user read a private repository's full config (format, blob store, remote proxy URL, `allowAnonymous`) just by knowing or guessing its name.
- **`PromotionHandler.ListRules`/`GetComponentRules`** — both sit under the non-admin `authed` group and return every promotion rule's full detail (source repo, target repo, path filter, approval gate) with no RBAC filtering — `ListRules` in particular returns *all* rules system-wide, not scoped to a single component.

**Where**
- `internal/api/handlers/scan.go` (`GetScanResult`)
- `internal/api/router.go` (`GET /service/rest/v1/assets/:id` inline handler)
- `internal/api/handlers/repositories.go` (`Get`)
- `internal/api/handlers/promotion.go` (`ListRules`, `GetComponentRules`)

**Impact**

Any authenticated user who knows or enumerates an id/name can read vulnerability reports, asset metadata, private repository config, or promotion-rule topology they have no privilege over.

**Suggested fix**

For each: resolve the owning repository and apply `rbacSvc.CanAccessRepo(...,"browse")` (or `FilterRepos`/a rules-visibility filter for the list case) before responding — 404 on denial, matching `ComponentHandler.Get`'s convention of keeping ids unguessable.

jorgemillan:
Closes #379

ComponentHandler.Get and RepositoryHandler.List both filter by RBAC (the former citing #291 explicitly), but four sibling read endpoints across the API didn't, letting any authenticated user read private data just by knowing or guessing an id or name:

- GET /api/v1/components/:id/scan (ScanHandler.GetScanResult) — full Trivy vulnerability report, CVE list included.
- GET /service/rest/v1/assets/:id (inline handler in router.go) — full asset metadata and download URL.
- GET /service/rest/v1/repositories/:name (RepositoryHandler.Get) — full repo config: format, blob store, remote proxy URL, allowAnonymous.
- GET /api/v1/promotion/rules and GET /api/v1/components/:id/promotion-rules (PromotionHandler.ListRules/GetComponentRules) — every rule's full topology (source/target repo, path filter, approval gate); ListRules in particular returns all rules system-wide, not scoped to one component.

All four now resolve the relevant repository (or, for promotion rules, both FromRepo and ToRepo) and check RBACService.CanAccessRepo/ FilterRepos/FilterComponents before responding — 404, not 403, on denial, so an id or name stays unguessable rather than confirming it exists. ScanHandler and PromotionHandler gained a nil-safe WithRBAC option (existing tests unaffected); RepositoryHandler already carried rbacSvc.

Verified: new unit tests for each of the three in-process handlers (TestScanHandler_GetScanResult_DeniedWithoutPrivilege, TestRepositoryHandler_Get_DeniedWithoutPrivilege,
TestPromotionHandler_ListRules_HidesRulesForReposWithoutPrivilege) and a new end-to-end integration test for the inline router.go handler (TestAssetGetByID_DeniedWithoutPrivilege, boots the real router over a real dockertest Postgres) — each confirms a privilege-less caller gets 404 and an admin still gets the real result. Live compose smoke test (real Postgres/MinIO) reproduced all four independently: created private repos/a promotion rule/a maven component as admin, confirmed a freshly created no-role user got 404 on all four endpoints while admin saw the real data, then cleaned up. go build/go vet/go test ./... (unit + -tags=integration) green save for the pre-existing, environment-only TestWebhookHandler_Test_200 flake (netguard blocking httptest.Server loopback in this sandbox, unrelated to this change).

### DIFF
--- a/internal/api/handlers/promotion.go
+++ b/internal/api/handlers/promotion.go
@@ -33,8 +33,12 @@ func (h *PromotionHandler) WithRBAC(repos repository.RepositoryRepo, rbacSvc *se
 	return h
 }
 
-// visibleRules filters rules to ones the caller can browse either FromRepo or
-// ToRepo. Nil-safe: with WithRBAC unset, returns rules unchanged.
+// visibleRules filters rules to ones the caller can browse both FromRepo and
+// ToRepo. A rule names both repositories, so browse on only one side would
+// still leak the other side's name, path filter, and approval gate — e.g.
+// someone with access to a "staging" repo could enumerate the production
+// repositories it promotes into. Nil-safe: with WithRBAC unset, returns rules
+// unchanged.
 func (h *PromotionHandler) visibleRules(c *gin.Context, rules []domain.PromotionRule) []domain.PromotionRule {
 	if h.repos == nil || h.rbacSvc == nil {
 		return rules
@@ -61,7 +65,7 @@ func (h *PromotionHandler) visibleRules(c *gin.Context, rules []domain.Promotion
 
 	visible := make([]domain.PromotionRule, 0, len(rules))
 	for _, r := range rules {
-		if canBrowse(r.FromRepo) || canBrowse(r.ToRepo) {
+		if canBrowse(r.FromRepo) && canBrowse(r.ToRepo) {
 			visible = append(visible, r)
 		}
 	}
--- a/internal/api/handlers/promotion.go
+++ b/internal/api/handlers/promotion.go
@@ -7,17 +7,65 @@ import (
 	"github.com/gin-gonic/gin"
 
 	"github.com/nexspence-oss/nexspence/internal/domain"
+	"github.com/nexspence-oss/nexspence/internal/repository"
 	"github.com/nexspence-oss/nexspence/internal/service"
 )
 
 // PromotionHandler serves the staging and build-promotion REST endpoints.
 type PromotionHandler struct {
-	svc *service.PromotionService
+	svc     *service.PromotionService
+	repos   repository.RepositoryRepo
+	rbacSvc *service.RBACService
 }
 
 // NewPromotionHandler constructs a PromotionHandler backed by the given promotion service.
 func NewPromotionHandler(svc *service.PromotionService) *PromotionHandler {
 	return &PromotionHandler{svc: svc}
+}
+
+// WithRBAC wires the repository repo and RBAC service so ListRules/
+// GetComponentRules hide rules naming a repository the caller has no browse
+// privilege over. Nil-safe: unset, both endpoints keep their prior
+// (unfiltered) behavior.
+func (h *PromotionHandler) WithRBAC(repos repository.RepositoryRepo, rbacSvc *service.RBACService) *PromotionHandler {
+	h.repos = repos
+	h.rbacSvc = rbacSvc
+	return h
+}
+
+// visibleRules filters rules to ones the caller can browse either FromRepo or
+// ToRepo. Nil-safe: with WithRBAC unset, returns rules unchanged.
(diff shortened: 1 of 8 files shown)


skensell201:
Good catch on all four — the 404-not-403 convention and the nil-safe `WithRBAC` options match how `ComponentHandler.Get` already does it, and the integration test booting the real router for the inline `assets/:id` handler is the right way to cover that one.

One thing to reconsider before merging: `visibleRules` keeps a rule when the caller can browse **either** `FromRepo` **or** `ToRepo`:

```go
if canBrowse(r.FromRepo) || canBrowse(r.ToRepo) {
```

A promotion rule names both repositories, so a caller with browse on only one side still learns the other side's name (plus the path filter and approval gate) — which is most of what the issue objects to leaking. Someone with access to a `staging` repo can enumerate the production repository names it promotes into.

`&&` would be the conservative reading: a rule is visible only to someone who can see both ends of it. That does hide rules from a caller who legitimately owns one side, so it's a judgment call rather than an obvious bug — but the `||` version should at least be a deliberate, commented decision rather than an incidental one, since it's the difference between "hides rules you have no business seeing" and "hides only rules where you have no business seeing *either* end."

Rest of the PR looks right. Verified locally alongside the other seven in this batch: `go build` / `go vet` clean, full suite green. Note it conflicts with #390 on one line of `internal/api/router.go` (the `replH` / `promotionH` constructor block) — trivial to resolve, both changes are compatible, but whichever lands second will need a rebase.

jorgemillan:
Went with && per your conservative reading — a rule is now only visible when the caller can browse both FromRepo and ToRepo. Added a test covering the partial-access case (browse on one side only) and verified live against the real API: a user with anonymous browse on a "staging" repo but no access to "production" gets an empty list instead of the rule.

skensell201:
Landed as #416 — merged into `main` in commit `230f9af`, with your commits and authorship intact.

The only change from this branch is the one-line conflict resolution in `internal/api/router.go`: #390 landed first and made it `NewReplicationHandler(replSvc, log)`, so the merged line keeps both that and your `NewPromotionHandler(promotionSvc).WithRBAC(repoRepo, rbacSvc)`. I opened it from a repo branch rather than asking you for a rebase because the whole batch was going out in one release — apologies for the detour past your PR.

Both review points are in: the `&&` reading of `visibleRules` here, and the pessimistic readiness default in #390. Thanks for the eight PRs — every one of them was a real bug, and the write-ups made them quick to verify. Closing this one as superseded.

Reviewed the full batch again on top of `main` before merging: `go build` / `go vet` / `golangci-lint` clean, `go test ./...` and the integration suite green.

jorgemillan:
Thanks for the patience and the great work.

PR #392, opened 2026-09-04
fix: S3 blob GC can abort an in-progress OCI upload session

### LINKED ISSUE #380: S3 blob GC can abort an in-progress OCI upload session that local storage would not
An OCI chunked upload session starts with a real empty object at the plain blob key (`oci.uploadStore.create` → a plain `BlobStore.Put`), separate from whatever `AppendBlob`'s own multipart machinery does later. On `LocalBlobStore`, every subsequent chunk reopens the real file with `O_APPEND`, so the file's mtime keeps advancing with real activity. On `S3BlobStore`, chunks go to multipart parts and a `.append-meta` side-object — neither touches that original plain-key placeholder, so `ListEntries`' `ModTime` for it stays frozen at session start, no matter how recently a chunk landed. GC's `min_age` grace period reads exactly that `ModTime` (and `ListEntries` explicitly *excludes* `.append-meta` from its scan, so the more-recent timestamp is invisible to it) — a long upload with a slow client, or an admin running GC with a shortened `min_age` to free space urgently, can abort (`AbortAppend`) a session that is still actively receiving data, purely because it has been open a while.

**Where**
- `internal/storage/s3.go` (`ListEntries`) — reports the placeholder object's own `ModTime`, never the `.append-meta` object's (which *is* rewritten on every `saveAppendState`)
- `internal/formats/oci/uploads.go` (`create`) — the plain-key `Put` that establishes the placeholder
- `internal/storage/s3_append.go` (`saveAppendState`) — writes `.append-meta` on every flushed part

**Impact**

Backend-specific data loss: an OCI push large/slow enough to outlive `min_age` gets silently aborted on S3/MinIO but not on local storage — asymmetric behavior between backends, worse the more aggressively an operator shortens `min_age` to reclaim space.

**Suggested fix**

Have `ListEntries` correlate each key's `.append-meta` object (when present) and use its `ModTime` if more recent than the placeholder's own.

jorgemillan:
Closes #380

An OCI chunked upload session starts with a real empty object at the plain blob key (oci.uploadStore.create -> a plain BlobStore.Put), separate from whatever AppendBlob's own multipart machinery does later. On LocalBlobStore, every subsequent chunk reopens the real file with O_APPEND, so the file's mtime keeps advancing with real activity. On S3BlobStore, chunks go to multipart parts and a .append-meta side-object - neither touches that original plain-key placeholder, so ListEntries' ModTime for it stayed frozen at session start no matter how recently a chunk landed (ListEntries explicitly excludes .append-meta from its scan, so the more-recent timestamp was invisible to it). GC's min_age grace period reads exactly that ModTime - a long upload with a slow client, or an admin running GC with a shortened min_age to free space urgently, could abort (via Delete -> AbortAppend) a session that was still actively receiving data, purely because it had been open a while. Backend-specific data loss: the same upload outlives min_age silently on S3/MinIO but not on local storage.

ListEntries now collects each key's .append-meta ModTime into a map during the same listing pass (still excluded from the returned entries themselves) and overrides an entry's ModTime with it when newer - a page can split a placeholder from its .append-meta companion (listing order is lexicographic, not paired), so the correlation only happens once every page has been seen.

Verified against a real MinIO (dockertest): a new integration test (TestS3BlobStore_ListEntries_TracksMostRecentAppendActivity) mirrors the real create()+AppendBlob sequence - a plain Put, a 1.2s wait, then a real chunk - and asserts ListEntries' ModTime tracks the chunk, not session start. Confirmed it fails without the fix (0s gap, ModTime pinned to session start) and passes with it. Full storage package suite green (go test -tags=integration ./internal/storage/...), including the existing TestS3BlobStore_AppendState_HiddenFromListings regression check and TestS3BlobStore_ListKeys (unaffected - it doesn't correlate .append-meta, only excludes it, same as before).

### DIFF
--- a/internal/storage/s3.go
+++ b/internal/storage/s3.go
@@ -207,9 +207,29 @@ func (s *S3BlobStore) ListKeys(ctx context.Context) ([]string, error) {
 	return keys, nil
 }
 
+// blobKeyFromObjectKey reverses objectKey's two-level sharding:
+// "ab/cd/abcdef..." → "abcdef...".
+func blobKeyFromObjectKey(objKey string) string {
+	parts := strings.SplitN(objKey, "/", 3)
+	if len(parts) == 3 {
+		return parts[2]
+	}
+	return objKey
+}
+
 // ListEntries returns every object's blob key, size and last-modified time.
 func (s *S3BlobStore) ListEntries(ctx context.Context) ([]BlobEntry, error) {
 	var entries []BlobEntry
+	// A chunked OCI upload session's real activity — every AppendBlob call —
+	// only ever rewrites the .append-meta side-object (saveAppendState),
+	// never the plain placeholder Put that create() makes at session start.
+	// Left alone, an entry's ModTime stays pinned to session start no matter
+	// how recently a chunk landed, and GC's min_age grace period can abort a
+	// session that is still actively receiving data. Track each key's most
+	// recent .append-meta ModTime here (still excluded from the returned
+	// entries themselves) and use it below when it postdates the
+	// placeholder's own.
+	metaModTimes := make(map[string]time.Time)
 	paginator := s3.NewListObjectsV2Paginator(s.client, &s3.ListObjectsV2Input{
 		Bucket: aws.String(s.bucket),
 	})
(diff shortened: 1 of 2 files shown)


skensell201:

Issue #336, opened 2026-08-26
CVE scan finds nothing for crates pulled through a Cargo proxy repository


UmanGarbag:
## Steps to reproduce

Create a Cargo repository, type proxy, upstream https://index.crates.io
Add a known-vulnerable crate to a test project and build through the proxy:

```toml
   [dependencies]
   smallvec = "=1.6.1"   # CVE-2021-25900 / RUSTSEC-2021-0003
Check the cached component:
   GET /service/rest/v1/search?repository=<repo>&name=smallvec

```
Result:

```json
   {
     "name": "sm/al/smallvec",
     "version": "1"
   }
```

(expected: "name": "smallvec", "version": "1.6.1") 

Run the scan:   `POST /api/v1/components/<id>/scan`

Response: "status": "ok", but summary.total: 0 — no vulnerability found.

## Expected

The scan should report CVE-2021-25900 for smallvec 1.6.1.

## Actual

The component is stored with a mangled name (full asset path) and a placeholder version ("1"), so the OSV query never matches the real package. Reproduces for every crate observed in the cache, not just this one the scan itself runs fine, it's just querying OSV with the wrong name/version.

So we don't have the possibility to have vulnerability for cargo and other repository

jorgemillan:
Reproduced and confirmed via code inspection + a live test against a local proxy build (reverted afterward, nothing committed).

## Root cause

`serveFile`'s proxy branch (`internal/formats/pypi/handler.go`) derives the cached component's coordinates from the *directory* of the requested path:

```go
pkgGuess := path.Base(path.Dir(filePath))
coords := base.Coords{Name: normalizePackageName(pkgGuess), Version: "wheel"}
```

`Version` is a literal `"wheel"` — never the real version, for any file. And real `files.pythonhosted.org` URLs are hash-bucketed (`/packages/<hash>/<hash>/<hash>/<file>`), not package-name paths, so `pkgGuess` isn't the package name either — it's a hash fragment. Both `comp.Name` and `comp.Version` are wrong by construction, which is why `scanOSV`'s `OSVClient.Query(ctx, comp.Name, comp.Version, ecosystem)` can never match anything.

## Same bug in Cargo

[...]

PR #393, opened 2026-09-04
fix: Yum and NuGet proxies never derive real package coordinates

### LINKED ISSUE #381: Yum and NuGet proxies never derive real package coordinates — invisible to vulnerability scanning
Same root cause `#336` closed for Cargo, in two more formats: both proxy branches pass an empty `base.Coords{}` for *everything* they cache from upstream, including real downloads — even though each format's hosted-upload path already parses real coordinates (Yum: NVR from the filename; NuGet: the trivially-parseable `/v3/flatcontainer/:id/:ver/:id.:ver.nupkg` path, same split the hosted download path already does). A cached package registers under a path-derived name and placeholder version, so the OSV/Trivy scan can never match it to a real advisory.

**Where**
- `internal/formats/yum/handler.go` — `coords := base.Coords{}` in the proxy branch; `handleUpload`'s existing NVR parser is not reused there
- `internal/formats/nuget/handler.go` — `coords := base.Coords{}` in the proxy branch; `serveFlatContainerDownload` (hosted) already does the split that would work here

**Impact**

Every RPM/package served through a Yum or NuGet proxy repository is invisible to vulnerability scanning, silently — no error, no operator-visible signal.

**Suggested fix**

A `proxyCoords(p)` per format reusing the existing parser/split already used on the hosted-upload path; `repodata/*`/registration-index pages keep the path-based metadata fallback.

jorgemillan:
Closes #381

Same root cause #336 closed for Cargo, in two more formats: both proxy branches passed an empty base.Coords{} for everything cached from upstream, including real downloads, even though each format's hosted-upload path already parses real coordinates. A cached artifact registered under a path-derived name and placeholder version, so OSV/Trivy scanning could never match it to a real advisory.

Yum: the NVR (name-version-release) parser inside handleUpload is now a shared rpmCoords(filename), used by a new proxyCoords(p) (any .rpm suffix) for the proxy branch too; repodata/* index files keep the path-based metadata fallback.

NuGet: a new proxyCoords(p) recognizes any .nupkg path by suffix and splits its last 3 segments (:id/:ver/:id.:ver.nupkg) - not the "/v3/flatcontainer/" prefix this handler's own hosted routes use locally. A proxy repo forwards whatever local path the client requested straight onto upstream, and a real client (nuget.exe, dotnet) requests packages at whatever address the upstream's index.json/config.json actually advertised - "/v3-flatcontainer/" (hyphenated) for nuget.org and most real feeds, confirmed against the existing TestNuGet_ProxyFlatcontainer_ResolvesAgainstRealShape test. Matching by suffix instead of a fixed prefix handles both shapes. Registration/index pages keep the metadata fallback.

Verified: new tests per format confirm a proxied download registers under its real coordinates and a metadata/index page keeps the fallback (TestYum_ProxyDownload_RegistersRealRPMCoords, TestYum_ProxyRepomd_KeepsMetadataFallback,
TestNuGet_ProxyDownload_RegistersRealPackageCoords - and a second NuGet test confirming the slash convention also still resolves, TestNuGet_ProxyRegistrationIndex_KeepsMetadataFallback). Live compose verification against real upstreams: proxied openssl3-3.5.5-6.el8 from a real Fedora EPEL 8 mirror through a live Yum proxy repo, and newtonsoft.json 13.0.1 from api.nuget.org (real /v3-flatcontainer/ shape) through a live NuGet proxy repo - both registered with their real name/version (not path-derived placeholders), confirmed via the API and visually in Chrome's Search page. go build/go vet/go test ./... green save for the pre-existing, environment-only TestWebhookHandler_Test_200 flake.

### DIFF
--- a/internal/formats/nuget/handler.go
+++ b/internal/formats/nuget/handler.go
@@ -65,7 +65,7 @@ func (h *Handler) ServeHTTP(c *gin.Context) {
 		if !strings.HasSuffix(p, ".nupkg") {
 			maxAge = repoproxy.MetadataMaxAge(repo)
 		}
-		coords := base.Coords{}
+		coords := proxyCoords(p)
 		// Resource paths advertised by the rewritten index are the upstream's
 		// own paths re-rooted locally — forward them onto the bare origin,
 		// not onto a possibly /v3-suffixed remote_url (#349).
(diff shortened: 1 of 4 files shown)


skensell201:

PR #394, opened 2026-09-04
fix: unbounded read of rewritten proxy metadata (npm/PyPI/NuGet) - me…

### LINKED ISSUE #382: Unbounded read of rewritten proxy metadata (npm/PyPI/NuGet) allows a memory DoS
`repoproxy.storeAndServeResponse` and `serveCachedAsset` — the shared path npm (packument), PyPI (simple index) and NuGet (registration/flatcontainer index) proxy through whenever they need to rewrite a response before serving it — do a bare `io.ReadAll` with no `io.LimitReader`, both when populating the cache from upstream and when re-reading the cache to rewrite it again. Every other metadata read in this codebase (OCI manifests/referrers, npm's own age-filter packument read, PyPI's simple index reader, RubyGems gemspecs) already caps its `io.ReadAll` with `io.LimitReader` and a dedicated constant — this is the one unbounded read, reachable from any proxy repository whose `remote_url` a caller can point at a hostile, compromised, or simply oversized-by-accident upstream.

**Where**
- `internal/formats/repoproxy/repoproxy.go` — `storeAndServeResponse` (upstream fetch) and `serveCachedAsset` (cache re-read), both gated on `rewrite != nil`

**Impact**

A proxy repository pointed at an upstream that returns an oversized response for a rewritten-metadata path exhausts server memory before any quota/size check runs — no additional privilege needed beyond "a proxy repo exists and its upstream is reachable."

**Suggested fix**

Wrap both reads in `io.LimitReader` with a shared cap, rejecting (not truncating) a response that exceeds it.

jorgemillan:
Closes #382

storeAndServeResponse and serveCachedAsset - the shared path npm (packument), PyPI (simple index) and NuGet (registration/flatcontainer index) proxy through whenever they need to rewrite a response before serving it - did a bare io.ReadAll with no io.LimitReader, both when populating the cache from upstream and when re-reading the cache to rewrite it again. Every other metadata read in this codebase (OCI manifests/referrers, npm's own age-filter packument read, PyPI's simple index reader, RubyGems gemspecs) already caps its io.ReadAll with io.LimitReader and a dedicated constant - this was the one unbounded read, reachable from any proxy repository whose remote_url a caller can point at a hostile, compromised, or simply oversized-by-accident upstream.

Added maxRewrittenMetadataBytes = 32 << 20 and wrapped both reads with io.LimitReader(reader, cap+1) - the same "+1" trick oci/handler.go and rubygems/handler.go already use, so a reader capped at exactly the limit can't silently return a trimmed prefix with a nil error instead of a rejection. Both now return an explicit error (502 from the format handlers, which already wrap ServeGET/ServeGETRewritten errors that way) instead of buffering past the cap.

Verified: new tests confirm a body one byte over the cap is rejected without being cached (storeAndServeResponse) and without being served (serveCachedAsset, seeded directly to bypass the fetch-time guard and exercise the cache re-read path independently), and a body exactly at the cap is still served normally. Confirmed both new tests fail without the fix (nil error / 200 where an error / 502 was expected) and pass with it. go build/go vet/go test ./... green. Live compose verification: a real npm proxy against registry.npmjs.org still serves a real packument (lodash, 117 versions, 241.8 KB) correctly under the new cap, confirmed via the API and visually in Chrome's Search page.

### DIFF
--- a/internal/formats/repoproxy/repoproxy.go
+++ b/internal/formats/repoproxy/repoproxy.go
@@ -174,6 +174,18 @@ func applyChecksumHeaders(c *gin.Context, a *domain.Asset) {
 // addressed by digest) are never revalidated — callers pass maxAge == 0 for those.
 const DefaultMetadataMaxAge = 10 * time.Minute
 
+// maxRewrittenMetadataBytes caps storeAndServeResponse's and
+// serveCachedAsset's buffered io.ReadAll — the only unbounded reads in this
+// codebase, both gated on rewrite != nil (npm packuments, PyPI simple pages,
+// NuGet registration/flatcontainer indexes: the shared path any proxy that
+// rewrites a response before serving it goes through). Every other metadata
+// read here (OCI manifests/referrers, npm's own age-filter packument read,
+// PyPI's simple index reader, RubyGems gemspecs) already caps its ReadAll
+// with a LimitReader; this was the one exception, reachable from any proxy
+// repository whose remote_url a caller can point at a hostile, compromised,
+// or simply oversized-by-accident upstream.
+const maxRewrittenMetadataBytes = 32 << 20
+
 // MetadataMaxAge returns the freshness TTL to use for proxied metadata on this
 // repository. It reads proxy_config["metadata_max_age"], interpreted as a number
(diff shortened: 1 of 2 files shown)


skensell201:

Issue #337, opened 2026-08-26
Four independent stale-state/race bugs in BrowsePage and SecurityPage (frontend)


jorgemillan:
## Summary

Four independent frontend bugs sharing the same failure class — component state that goes stale across an async boundary (a fetch, a race, an unmount) with no guard to catch it. All four were re-verified today (2026-08-26) end-to-end against a running instance, driving a real Chrome browser via Playwright — not just read from the source. Details, live evidence, and suggested fixes below, ordered by severity.

---

### 1. BrowsePage's "Promote" action can silently apply the wrong rule (security-relevant)

**Where:** `frontend/src/pages/BrowsePage.tsx:1482-1492` (Docker-tag handler), `:1606-1616` (raw-file handler)

Both single-item "Promote" handlers do:
```js
const res = await apiClient.get(`/api/v1/components/${id}/promotion-rules`)
setPromotionRules(res.data)
...
setSelectedRuleID(promotionRules[0]?.id ?? '')   // reads the closure's PREVIOUS `promotionRules`, not res.data
```
`promotionRules` in that last line is still the value from *before* this fetch (React state updates are async) — not the list that was just fetched. The bulk multi-select handler resets to `''` correctly; only these two single-item paths have the bug.

**Live repro (Chrome, today):** opened Promote on a raw file in repo A (2 applicable rules), selected rule "rule-strict", cancelled without submitting. Switched to a *different* repo/component (B) via the in-app repository picker — no full page reload — where only one, unrelated rule applies. Opened Promote again:
- The rule dropdown shows the placeholder **"Select a rule"** (rule-strict's id doesn't match repo B's rule, so nothing visibly matches).
- The **Promote submit button is NOT disabled** (`disabled={!selectedRuleID}` — `selectedRuleID` is still rule-strict's real, non-empty id internally).

An operator sees an apparently-empty selection and can still click "Promote," silently submitting a rule from a completely unrelated repo pair against this component.

**Suggested fix:** seed `selectedRuleID` from the freshly-fetched response (`const rules = res.data; setPromotionRules(rules); setSelectedRuleID(rules[0]?.id ?? '')`), and/or reset `selectedRuleID` to `''` before the fetch at both call sites, matching the bulk handler.

---

### 2. Vulnerability Dashboard's repo filter shows stale results

**Where:** `frontend/src/pages/SecurityPage.tsx:749-771` (`loadVulns`, the `useEffect` firing it on every `repoFilter` change)

[...]

PR #395, opened 2026-09-04
fix: frontend stale-response races in SecurityPage roles and connection tests

### LINKED ISSUE #383: Frontend: several pages are missing a request-sequence guard against late/stale responses
Two instances of the same missing pattern — `#337` already fixed this exact bug class in `VulnDashTab` (`vulnReqSeq`, a `useRef` sequence guard), left open in two sibling spots:

- **`SecurityPage.tsx`'s `openEdit(role)`** fires `Promise.all([listPrivileges(), listRolePrivileges(role.id)])` with no sequence guard. Opening "Edit" on role A, then closing/opening "Edit" on role B before A's slower request resolves, lets A's late response land after B's fast one and silently overwrite the checklist the admin is now looking at (and would save) for B. This one is a real security-relevant race, not just a UX glitch.
- **`AdminPage.tsx`'s `CreateBlobStoreModal`** shows a "✓ Connection successful" banner after `handleTest`, but none of `bucket`/`region`/`endpoint`/`accessKey`/`secretKey`/`path`'s `onChange` handlers clear `testResult` — editing any of them after a successful test leaves the stale green banner describing a configuration that was never tested, and a late in-flight test response can also land after fields already changed. `MigrationPage.tsx`'s equivalent `handleTest` does clear its preview state on edit (better) but has the same missing race guard for a late response.

**Where**
- `frontend/src/pages/SecurityPage.tsx` (`openEdit`)
- `frontend/src/pages/AdminPage.tsx` (`CreateBlobStoreModal.handleTest`, field `onChange` handlers)
- `frontend/src/pages/MigrationPage.tsx` (`handleTest`)

**Impact**

`SecurityPage`: silent corruption of RBAC configuration — a role can end up saved with a completely different role's privilege set, no error shown. `AdminPage`/`MigrationPage`: an admin can create a blob store believing its credentials/endpoint were verified when they were not.

**Suggested fix**

Same `useRef` sequence-number guard `vulnReqSeq` already demonstrates in `SecurityPage.tsx`'s own file, applied to `openEdit` (checked before applying the response, bumped on cancel too) and to both `handleTest`s; `AdminPage.tsx` also needs `testResult` cleared on every relevant field's `onChange`.

jorgemillan:
Closes #383

Two missing request-sequence-guard bugs in the same class #337 already fixed once (VulnDashTab's vulnReqSeq, in SecurityPage.tsx itself):

- SecurityPage.tsx's openEdit(role) fired Promise.all([listPrivileges(), listRolePrivileges(role.id)]) with no sequence guard. Opening Edit on role A, then closing/opening Edit on role B before A's slower request resolved, let A's late response land after B's fast one and silently overwrite the privilege checklist the admin was looking at (and would save) for B. Added editReqSeq (useRef), bumped at the start of every openEdit and on Cancel, checked before applying the response.

- AdminPage.tsx's CreateBlobStoreModal showed a "Connection successful" banner after handleTest, but none of the S3/path field onChange handlers cleared testResult - editing any of them after a successful test left the stale green banner describing a config that was never actually tested. The fields also weren't disabled during the test, so a fast edit could race a still-in-flight test's late response landing after the fields already changed. MigrationPage.tsx's equivalent handleTest already cleared its preview state on edit (better than AdminPage) but had the same missing race guard. Added a testReqSeq ref to both, bumped by every relevant field's onChange (also clearing the result/preview immediately) and by handleTest/the test itself at start; the response handler discards a reply whose seq no longer matches current. testBusy/testing is cleared in every case regardless of seq - nothing else ever resets it for a superseded request, so gating it the same way as the result would leave the button stuck showing "Testing..." forever after an edit invalidated an in-flight test (caught live in Chrome while verifying the first version of this fix, see below).

Verified live in a real Chrome browser (Playwright) against the running app (docker compose build, real tsc/vite build):
- SecurityPage: delayed the "Acceso Anonimo" role's privilege fetch via network interception, opened its Edit, immediately cancelled and edited "DAC" instead - the modal correctly showed "Edit Role: DAC" / 0 selected privileges both immediately and after the delayed Acceso Anonimo response landed (screenshots at both points confirm no corruption).
- AdminPage: ran a real local-filesystem connection test to success, edited the Path field, confirmed the banner vanished immediately; then delayed a second test via network interception, edited the field again while it was in flight, and confirmed after the delayed response landed that no stale "successful" banner appeared and the button returned to "Test Connection" (not stuck on "Testing..."). No console errors across the sweep.

### DIFF
--- a/frontend/src/pages/AdminPage.tsx
+++ b/frontend/src/pages/AdminPage.tsx
@@ -1915,6 +1915,12 @@ function CreateBlobStoreModal({ blobStores, onClose }: { blobStores: BlobStore[]
   const [err, setErr] = useState('')
   const [testResult, setTestResult] = useState<{ ok: boolean; error?: string } | null>(null)
   const [testBusy, setTestBusy] = useState(false)
+  // Bumped by handleTest and by every field onChange that feeds its config.
+  // A field edited after a test started must not let that test's late
+  // response land as if it verified the (now different) config a moment
+  // later — same request-sequence-guard pattern issue 12 applies to
+  // SecurityPage.tsx's openEdit.
+  const testReqSeq = useRef(0)
   const [groupFillPolicy, setGroupFillPolicy] = useState<'round_robin' | 'write_to_first_fill'>('round_robin')
   const [groupMemberIds, setGroupMemberIds] = useState<string[]>([])
 
@@ -1945,6 +1951,7 @@ function CreateBlobStoreModal({ blobStores, onClose }: { blobStores: BlobStore[]
   })
 
   const handleTest = async () => {
+    const seq = ++testReqSeq.current
     setTestBusy(true)
     setTestResult(null)
     try {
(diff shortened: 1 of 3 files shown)


skensell201:

PR #396, opened 2026-09-04
fix: Update handlers skip the field validation Create enforces

### LINKED ISSUE #384: `Update` handlers skip the field validation their `Create` sibling enforces (Replication, Promotion)
Two instances of the same pattern:

- **`ReplicationHandler.Update`** — `Create` requires `name`/`source_repo`/`target_url`/`target_repo` non-empty; `Update` has no such check, so a client can blank out `target_url`/`source_repo` on an existing rule — the cron then fails every tick (`unsupported protocol scheme ""`) with no signal at update time.
- **`PromotionService.UpdateRule`** — `CreateRule` rejects an empty `FromRepo`/`ToRepo`; `UpdateRule` only checks `FromRepo == ToRepo` (which also happens to catch the *both-empty* case, since `"" == ""`, but not "one side blank"), so `PUT /api/v1/promotion/rules/:id` with `from_repo:""` can persist an orphaned rule. Admin-only, hence lower severity than the Replication one.

**Where**
- `internal/api/handlers/replication.go` (`Update`)
- `internal/service/promotion_service.go` (`UpdateRule`)

**Suggested fix**

Mirror each `Create`'s validation in its `Update` sibling.

jorgemillan:
Closes #384

Two instances of the same "Update skips Create's validation" pattern:

- ReplicationHandler.Update: Create requires name/source_repo/ target_url/target_repo non-empty; Update had no such check, so a client could blank out target_url/source_repo on an existing rule - the cron then fails every tick (unsupported protocol scheme "") with no signal at update time. Mirrored Create's validation into Update.

- PromotionService.UpdateRule: CreateRule rejects an empty FromRepo/ ToRepo; UpdateRule only checked FromRepo == ToRepo (which also happens to catch the both-empty case, since "" == "", but not "one side blank"), so PUT /api/v1/promotion/rules/:id with from_repo:"" could persist an orphaned rule. Admin-only endpoint, hence low severity. Added the same non-empty check CreateRule has.

Verified: new tests mirror each Create-side test onto Update/UpdateRule (TestReplicationHandler_Update_MissingFields_400, and two new subtests in TestPromotionService_UpdateRule_Validation for a blank FromRepo/ ToRepo). Confirmed both fail without the fix (200 instead of 400 / nil error instead of the expected message) and pass with it. go build/go vet/go test ./... green. Live compose verification: created a real replication rule and a real promotion rule, confirmed blanking target_url (replication) and from_repo (promotion) via their Update endpoints now correctly returns 400 with the same message Create uses, and that the existing rules still render correctly in Chrome's Replication and Promotion tabs with zero console errors.

### DIFF
--- a/internal/api/handlers/replication.go
+++ b/internal/api/handlers/replication.go
@@ -86,6 +86,13 @@ func (h *ReplicationHandler) Update(c *gin.Context) {
 		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
 	}
+	// Mirrors Create's validation: without it, a client could blank out
+	// target_url/source_repo on an existing rule — the cron then fails every
+	// tick (unsupported protocol scheme "") with no signal at update time.
+	if inp.Name == "" || inp.SourceRepo == "" || inp.TargetURL == "" || inp.TargetRepo == "" {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "name, source_repo, target_url, target_repo are required"})
+		return
+	}
 	rule := &domain.ReplicationRule{
 		ID:             c.Param("id"),
 		Name:           inp.Name,
--- a/internal/api/handlers/replication_test.go
+++ b/internal/api/handlers/replication_test.go
@@ -134,6 +134,24 @@ func TestReplicationHandler_Update_BadJSON_400(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, rec.Code)
 }
 
+// Without this check, a client could blank out target_url/source_repo on an
+// existing rule — the cron then fails every tick with no signal at update time.
+func TestReplicationHandler_Update_MissingFields_400(t *testing.T) {
+	r := mountReplication(t)
+	id := replicationCreate(t, r, "before")
+	// each row drops exactly one required field
+	cases := []map[string]any{
+		{"source_repo": "src", "target_url": "http://x/", "target_repo": "dst"}, // no name
+		{"name": "n", "target_url": "http://x/", "target_repo": "dst"},          // no source_repo
+		{"name": "n", "source_repo": "src", "target_repo": "dst"},               // no target_url
+		{"name": "n", "source_repo": "src", "target_url": "http://x/"},          // no target_repo
+	}
+	for i, body := range cases {
+		rec := do(t, r, http.MethodPut, "/api/v1/replication/rules/"+id, body)
+		assert.Equalf(t, http.StatusBadRequest, rec.Code, "case %d body=%s", i, rec.Body.String())
+	}
+}
+
 func TestReplicationHandler_Update_UnknownID_400(t *testing.T) {
 	r := mountReplication(t)
 	// mock UpdateRule returns an error for an unknown id → handler maps to 400.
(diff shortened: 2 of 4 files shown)


skensell201:
This is the same fix I had in flight as #397, and yours landed first with live verification behind it — I've closed mine in favour of this one. Both checks read correctly.

One gap worth folding in while you're here, same class as the two you found. `validateCronExpr` returns nil for an empty expression:

```go
func validateCronExpr(expr string) error {
	if expr == "" {
		return nil
	}
```

`Create` covers that by defaulting a missing `cron_expr` to `"0 2 * * *"`, but `Update` has no such default — so a `PUT` that simply omits `cron_expr` (the natural shape for a client editing only the target URL) persists a rule with no schedule at all, silently unscheduling it. Same failure mode as blanking `target_url`: accepted with a 200, no signal until someone notices the rule stopped running.

The default and the required-field check want to move together, which is what pushed me toward a shared helper on `ruleInput` rather than the literal duplicated in both handlers:

```go
// It writes the 400 itself and reports whether the caller may proceed.
func (inp *ruleInput) validate(c *gin.Context) bool {
	if inp.Name == "" || inp.SourceRepo == "" || inp.TargetURL == "" || inp.TargetRepo == "" {
		c.JSON(http.StatusBadRequest, gin.H{"error": "name, source_repo, target_url, target_repo are required"})
		return false
	}
	if inp.CronExpr == "" {
		inp.CronExpr = defaultReplicationCron
	}
	return true
}
```

With a test alongside the one you already added:

```go
func TestReplicationHandler_Update_DefaultsMissingCronExpr(t *testing.T) {
	r := mountReplication(t)
	id := replicationCreate(t, r, "before")
	rec := do(t, r, http.MethodPut, "/api/v1/replication/rules/"+id, map[string]any{
		"name": "after", "source_repo": "src",
		"target_url": "http://127.0.0.1:1/", "target_repo": "dst",
	})
	require.Equal(t, http.StatusOK, rec.Code, "body=%s", rec.Body.String())
	var rule domain.ReplicationRule
	require.NoError(t, json.Unmarshal(rec.Body.Bytes(), &rule))
	assert.Equal(t, "0 2 * * *", rule.CronExpr)
}
```

It fails on current `main` with `expected "0 2 * * *", actual ""`. Take it or leave it as a follow-up — the rest of this PR is good either way.

skensell201:
My mistake, and I owe you a clear account of it.

I told you above I'd closed #397 in favour of this PR. What I missed is that the branch behind #398 (the OCI streaming fix) had been cut on top of the #384 commit rather than on `main`, so #398 carried two commits, not one. Merging it a few minutes ago landed the #384 fix on `main` as a side effect and auto-closed the issue. That was not a judgement that this PR wasn't good enough — it was a branching error on my side, and I should have checked #398's commit list before opening it.

What is on `main` now (515a575) is the version I'd described here as a suggestion: the shared `ruleInput.validate` helper plus `Create`'s `cron_expr` default applied on update, with tests for both. So it is a superset of this PR rather than a divergence from it, and the finding is yours either way — #384 is your issue.

I'd rather not leave this open to rot, so I'm closing it. If you'd prefer your version to be the one in history, say so and I'll revert 515a575's replication/promotion hunks and merge this PR instead — no argument from me, it's a fair ask.

The other seven in your batch are unaffected and still waiting on the two review notes in #390 and #391.

jorgemillan:
Thanks for the honest account — no worries, branching mistakes happen. 515a575 on main already covers what this PR set out to fix, so no need to revert anything. Leaving #396 closed as-is.

PR #397, opened 2026-09-04
fix(replication,promotion): apply Create's field validation in the Update siblings

### LINKED ISSUE #384: `Update` handlers skip the field validation their `Create` sibling enforces (Replication, Promotion)

skensell201:
Closes #384

Two instances of the same pattern, where a `PUT` could persist a rule its own `POST` would have rejected.

**`ReplicationHandler.Update`** never ran `Create`'s required-field check, so blanking `target_url` or `source_repo` on an existing rule was accepted with a 200 and only surfaced afterwards, as the cron failing every tick with `unsupported protocol scheme ""`. Both handlers now share a `ruleInput.validate`. It also applies `Create`'s cron default: `validateCronExpr` accepts an empty expression, so an update that omitted `cron_expr` silently unscheduled the rule instead of leaving it on a schedule.

**`PromotionService.UpdateRule`** checked only `FromRepo == ToRepo`. That happens to catch a rule with *both* sides blank (`"" == ""`) but not one with a single side blank, so a `PUT` with `from_repo:""` persisted an orphaned rule no promotion could ever match. It now runs `CreateRule`'s own emptiness check first.

**Verified**: three new tests — `TestReplicationHandler_Update_RejectsBlankedRequiredFields` (each of the four fields blanked in turn), `TestReplicationHandler_Update_DefaultsMissingCronExpr`, and `TestPromotionService_UpdateRule_RejectsBlankRepo` (which also asserts the stored rule is untouched by a rejected update). Confirmed all three fail against the unpatched code (200 instead of 400, `""` instead of the default cron, `<nil>` instead of an error) and pass with it. `go build` / `go vet` / `go test ./...` green apart from the pre-existing sandbox-only `TestWebhookHandler_Test_200`.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UnmG6Yeh64hPJjqLGDnS6o

### DIFF
--- a/internal/api/handlers/replication.go
+++ b/internal/api/handlers/replication.go
@@ -44,6 +44,30 @@ type ruleInput struct {
 	Enabled        bool   `json:"enabled"`
 }
 
+// defaultReplicationCron is the schedule a rule falls back to when the caller
+// sends none: nightly at 02:00.
+const defaultReplicationCron = "0 2 * * *"
+
+// validate rejects an input that would persist a rule the replication cron
+// cannot actually run, and fills in the default schedule. Update needs exactly
+// the same guard Create does: a PUT that blanks target_url or source_repo used
+// to be accepted, and the only signal was the cron failing every tick after
+// the fact ("unsupported protocol scheme \"\""). An empty cron_expr is likewise
+// accepted by validateCronExpr, so leaving it blank on update silently
+// unscheduled the rule instead of keeping it on its old schedule.
+//
+// It writes the 400 itself and reports whether the caller may proceed.
+func (inp *ruleInput) validate(c *gin.Context) bool {
+	if inp.Name == "" || inp.SourceRepo == "" || inp.TargetURL == "" || inp.TargetRepo == "" {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "name, source_repo, target_url, target_repo are required"})
+		return false
+	}
+	if inp.CronExpr == "" {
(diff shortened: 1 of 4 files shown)


skensell201:
Closing in favour of #396, which fixes the same issue and was opened first (with live verification). I've moved the one extra finding — Update not applying Create's `cron_expr` default, which silently unschedules a rule on a PUT that omits it — over to that PR as a suggested follow-up. Branch `fix/update-handler-validation` deleted.

PR #415, opened 2026-09-04
feat(helm)!: bundle PostgreSQL instead of the Bitnami sub-chart

n00bi14:
## Why

The Helm chart depended on `bitnami/postgresql` 15.5.38, vendored under `charts/`
as a 74 KiB tarball plus a `Chart.lock`. That dependency has become a liability:
Bitnami removed the legacy `docker.io/bitnami/*` tags from Docker Hub in 2025 and
moved the maintained images into the paid `bitnamisecure` catalog. Only frozen
`bitnamilegacy/*` copies remain, so a default install pinned an image nobody
patches any more — and every install snippet in our docs began with a
`helm dependency update` that needed network access to Docker Hub.

For a single StatefulSet, owning ~150 lines of template we can read and pin is
cheaper than depending on a vendor that just changed its distribution terms.

## What changed

**Bundled database (`postgresql.enabled=true`, the default)** — the chart now
renders its own Service, StatefulSet, password Secret, probes, security context
and PVC from `docker.io/library/postgres:18`. Still a convenience database:
single replica, no replication, no backups, and deliberately no
PodDisruptionBudget. Production installs should keep setting
`postgresql.enabled=false`.

A few details are load-bearing and easy to "clean up" by accident:

- Resources are named `{release}-postgres`, **not** the Bitnami
  `{release}-postgresql`. Helm would otherwise try to patch the old StatefulSet
  in place, which fails on the immutable `spec.selector` and
  `volumeClaimTemplates`. The distinct name makes this a create and leaves the
  Bitnami PVC in place as a `pg_dump` source.
- `PGDATA` is left at the image default (a subdirectory of the mounted volume);
  `initdb` refuses to run on a PVC root that already holds `lost+found`.
- The DSN no longer carries a password. The app receives it via `PGPASSWORD`,
  which pgx honours, so the password never appears in a rendered manifest.
- A `preStop` hook runs `pg_ctl -m fast`: Kubernetes sends SIGTERM (smart
  shutdown) and ignores the image's SIGINT `STOPSIGNAL`, so open connections
  would otherwise hold the process until SIGKILL and crash recovery.
- The app waits on `pg_isready` in an initContainer instead of crash-looping
  through its own connection retries.

**Secret-based configuration** — `externalDatabase.existingSecret` (key `dsn`,
overridable) lets the external DSN stay out of values files entirely; when set,
the chart doesn't compose `NEXSPENCE_DATABASE_DSN` at all and the Deployment
references the operator's Secret directly. `postgresql.auth.existingSecret` does
the same for the bundled password. Configuring neither DSN source now fails
during templating rather than at pod startup.

**`values.schema.json`** — Helm silently drops unknown values keys, so an
operator carrying a `postgresql.primary.persistence` block forward from the
Bitnami layout would get a default 10Gi PVC on the default storage class while
their file plainly says otherwise. `additionalProperties: false` over the
`postgresql` and `externalDatabase` trees turns that into an install-time error
naming the offending path.

**CI** — the chart previously had no coverage. A new job lints it, renders every
database configuration and validates the output against upstream Kubernetes
schemas with kubeconform. Assertions cover the regressions that are cheap to
reintroduce and expensive to notice: no password slot in the bundled DSN, no
`postgresql.auth.password` in a manifest when an existing Secret is set, a
username with a space rendering as `%20` (not the `+` that `urlquery` emits and
pgx reads literally), no StatefulSet when the bundled DB is disabled, no
`volumeClaimTemplates` when persistence is off, and all three guard rails
actually aborting. Every `values-examples/*.yaml` is rendered too, so the
published networking examples can't drift from the chart.

## Breaking changes

- The bundled database goes from PostgreSQL **16** (Bitnami) to **18** and is
  renamed `{release}-postgres`. A two-major jump means `PGDATA` cannot be
  upgraded in place.
- **`helm upgrade` is not a migration path** for an existing Bitnami install —
  neither the rename nor the data directory is handled automatically. Dump with
  `pg_dump -Fc` from the old StatefulSet, delete it (`volumeClaimTemplates` PVCs
  survive, so `data-nexspence-postgresql-0` stays available as a source),
  install this chart, then `pg_restore`. Full steps in the chart README.
- Persistence values move from `postgresql.primary.persistence` to
  `postgresql.persistence`. The schema rejects the old keys instead of ignoring
  them.
- Password rotation now works the way the official image works:
  `POSTGRES_PASSWORD` is only applied by `initdb` on an empty data directory, so
  changing the value later has no effect until the role itself is altered with
  `\password`. Documented in the chart README.
- `helm dependency update` must be dropped from install commands — with no
  dependencies left, it now fails.


skensell201:
Read the whole thing, merged it onto current `main` (it lands clean over #418 and #419, which also touched `values.yaml`, `configmap.yaml`, `secret.yaml` and `deployment.yaml`) and rendered every database configuration. The reasoning is right on every detail I checked, and several of them are the kind that only show up in production:

- `{release}-postgres`, not `-postgresql` — correct, and the reason (immutable `spec.selector` / `volumeClaimTemplates`) is exactly why an in-place patch would fail.
- Leaving `PGDATA` at the image default. `postgres:18` puts it at `/var/lib/postgresql/18/docker` under the `VOLUME`, and mounting the parent is what keeps `initdb` off a PVC root holding `lost+found`.
- uid/gid 999 with `fsGroup: 999` matches the image's `postgres` user, so `runAsNonRoot` actually works here.
- The passwordless DSN plus `PGPASSWORD`. pgx applies libpq environment defaults for fields the URI omits, so this works, and it keeps the credential out of every rendered manifest.
- The `%20`-not-`+` username encoding, and a CI assertion for it. That is a bug nobody finds twice.
- `preStop` with `pg_ctl -m fast` — Kubernetes does send SIGTERM and does ignore the image's `STOPSIGNAL`.
- `additionalProperties: false` over the `postgresql` and `externalDatabase` trees. Helm silently dropping a carried-forward `postgresql.primary.persistence` block is precisely the failure worth turning into a rendering error.

## One thing I'd like before merging

**`helm upgrade` from a Bitnami install comes up silently on an empty database.**

The README says an upgrade is not a migration path, but nothing stops one. The rename means the upgrade *succeeds*: a fresh StatefulSet initdb's an empty volume, the app's initContainer waits for it, `pg_isready` passes, and the server bootstraps a new admin against an empty schema. From the operator's side the instance looks wiped — no repositories, no users, `admin`/`admin123` back at the door — while the real data sits untouched in a `data-{release}-postgresql-0` PVC nobody has told them about. The failure is quiet and it lands in production, which is the worst pairing.

A `lookup` guard would close it — `lookup` returns an empty map during a dry run and on install, so it only fires on a real upgrade against a live cluster:

```
{{- if and .Values.postgresql.enabled (not .Values.postgresql.acknowledgeBitnamiMigration) }}
{{- $legacy := lookup "apps/v1" "StatefulSet" .Release.Namespace (printf "%s-postgresql" .Release.Name) }}
{{- if $legacy }}
{{- fail "This release still runs the Bitnami PostgreSQL StatefulSet. `helm upgrade` will NOT migrate its data: pg_dump first (see the chart README), then re-run with --set postgresql.acknowledgeBitnamiMigration=true." }}
{{- end }}
{{- end }}
```

The flag then becomes the thing the README's migration steps end with, and an operator who has already dumped types one extra `--set` instead of finding out afterwards.

## Merge timing

Holding this one out of today's release rather than asking you to rush the guard. `feat(helm)!` makes svu bump the whole product to `v3.0.0`, and a chart-only breaking change shouldn't be what decides that on its way past — the next release can carry it deliberately, with the migration steps in the release notes where an upgrader will actually see them.

Everything else in here is ready. Thanks for taking on the Bitnami problem — that vendored tarball has been a liability since they moved the images, and owning 150 lines of StatefulSet we can read is the right trade.

PR #423, opened 2026-09-07
feat(helm): close the gap between what the app supports and the chart exposes

### LINKED ISSUE #422: Helm chart: 11 gaps between what the app supports and what the chart can configure
## Context

Found while reviewing the chart for a real production deployment. Filing as one issue rather than eleven because most of these share the same shape — a feature (or a correct default) that already works at the application level has no `values.yaml` key to reach it — and reviewing them together makes the overlaps visible (issues 3/6 are the coordination/storage halves of the same "more than one replica" story; issue 1's fix removes the need to special-case issue 6's PVC; issue 11 came out of combining the rest — every credential the other ten introduce or touch needed the same `existingSecret` treatment).

Every fix below has been implemented and verified locally — individually, and then all eleven combined into one chart at once (`helm lint --strict` / `helm template` / `helm install --dry-run`, every toggle active simultaneously, including every credential sourced from an existing Secret rather than a plaintext value) — happy to open PRs for any/all of these if that's more useful than a values-schema discussion in one thread.

**Note on PR #415:** every new key below will need a corresponding entry in the new `values.schema.json` (`additionalProperties: false` would otherwise silently drop them) if #415 lands first.

---

## 1. `storage.type: s3` is non-functional — `NEXSPENCE_STORAGE_DEFAULT_TYPE` is never set (High)

`configmap.yaml` emits `NEXSPENCE_STORAGE_LOCAL_BASE_PATH` or `NEXSPENCE_STORAGE_S3_*` based on `.Values.storage.type`, but never sets `NEXSPENCE_STORAGE_DEFAULT_TYPE` itself — the one variable that tells the app which backend to use (`storage.default_type`, defaults to `"local"`). Setting `storage.type: s3` has zero effect; the app silently stays on local storage, which is then unwritable since the `blobs` PVC/volumeMount is gated on `storage.type == "local"` too.

**Fix:** add `NEXSPENCE_STORAGE_DEFAULT_TYPE: {{ .Values.storage.type | quote }}` to `configmap.yaml`. No change needed to the PVC gate — once `default_type` actually switches to `s3`, not mounting the local PVC is correct.

## 2. Liveness/readiness probes point at a dummy endpoint (High)

`livenessProbe`/`readinessProbe` both target `/service/rest/v1/status/check`, which unconditionally returns `200 {"status":"ok"}` — it checks nothing. The app has real probes at `/healthz` (process-alive) and `/readyz` (checks Postgres, and Redis when configured, with a short-TTL cache), unused by the chart. Kubernetes marks the pod Ready even when Postgres is completely unreachable.

**Fix:** point `livenessProbe` → `/healthz`, `readinessProbe` → `/readyz`.

## 3. No way to configure Redis — breaks distributed-lock coordination above one replica (High)

`RedisConfig` backs a distributed lock (serializes cleanup/GC/replication/migration cron runs across replicas) and the failed-login rate limiter's shared state. With no Redis configured, the app falls back to a no-op lock and per-instance rate limiting — silently, with no error. The chart has zero `redis` keys anywhere, yet fully supports `replicaCount > 1`/`autoscaling`.

**Fix:** a `redis:`/`externalRedis` values pair with the exact same dual shape #415 gives `postgresql`/`externalDatabase` — a bundled, single-replica `StatefulSet` (own template, not a Bitnami/vendor sub-chart) for the convenience case, and a plain `addr`/`password`/`db` block for pointing at a Redis run outside the chart:

- `redis.enabled` (default `true`, single-replica convenience instance — no clustering, no PodDisruptionBudget, same posture as the bundled Postgres)
- `redis.image.repository`/`tag` — official `redis`, not a vendor image
- `redis.auth.password` / `redis.auth.existingSecret` — same existing-Secret escape hatch `postgresql.auth` gets, so the password never has to sit in a values file
- `redis.persistence.enabled` (default `false` — this Redis only backs a distributed lock and the login rate limiter, both fine to lose on restart; unlike Postgres there's no durable data to protect) / `size` / `storageClass` for the case an operator wants it anyway
- `externalRedis.enabled` / `addr` / `password` / `db` for a Redis run outside the chart
- A `nexspence.redisInUse` helper (same idea as `postgresql`/`externalDatabase`) resolves whichever of the two is active and fails loudly if both are enabled at once, rather than silently picking one
- Call out in the `autoscaling`/`replicaCount` comment that one of the two Redis blocks (and a shared blob store, §6) is required above one replica

Template shape mirrors `templates/postgresql.yaml` from #415 closely: a headless Service + single-replica `StatefulSet` (`volumeClaimTemplates` only rendered when `persistence.enabled`, an `emptyDir` otherwise), the password delivered via `secretKeyRef` and read into `REDIS_PASSWORD`, `command: ["sh", "-c", "exec redis-server --requirepass \"$REDIS_PASSWORD\""]` (the official image takes the password as a CLI flag, not an env var the way Bitnami's did), and `livenessProbe`/`readinessProbe` running `redis-cli -a "$REDIS_PASSWORD" ping` instead of `pg_isready`.

## 4. No way to configure OpenTelemetry tracing (High)

`TracingConfig` is a complete, working feature (`docker-compose.yml`'s `tracing` profile already exercises it locally against a bundled Jaeger). The chart has zero `tracing` keys.

**Fix:** a `tracing:` values block mirroring `TracingConfig`'s fields (`enabled`, `otlpEndpoint`, `otlpProtocol`, `otlpInsecure`, `sampleRatio`, `serviceName`, `environment`), wired into `configmap.yaml` — none of these are secret.

## 5. No way to set `http.trusted_proxies` (High)

Every chart deployment sits behind an ingress/gateway (the chart supports nginx/Traefik/Istio/Cilium), yet `trusted_proxies` isn't exposed — left at its empty default, the app ignores `X-Forwarded-For` entirely, so the audit log and any IP-based logic only ever see the ingress pod's own IP.

**Fix:** `config.trustedProxies` (a list) in `values.yaml`, rendered as `NEXSPENCE_HTTP_TRUSTED_PROXIES` (comma-separated — confirmed against `TestLoad_TrustedProxies_EnvOverride`, viper's default decode hook already splits on `,`). Deliberately not defaulting it to the pod CIDR or `"*"` — that's an opinionated choice a chart maintainer should make explicitly, not bake in silently.

## 6. Local storage PVC is hardcoded to `ReadWriteOnce` (High)

No way to use a shared filesystem (e.g. EFS via `efs-csi`) instead of S3 for more than one replica — `LocalBlobStore`'s writes (atomic rename, single-writer append) don't depend on POSIX locking, so `ReadWriteMany` is safe, but the chart always creates its own `ReadWriteOnce` PVC with no bring-your-own-claim option.

**Fix:** `storage.local.accessMode` (default `ReadWriteOnce`, overridable to `ReadWriteMany`) and `storage.local.existingClaim` (when set, skip the chart's own PVC and mount that claim instead).

## 7. `NEXSPENCE_HTTP_LISTEN` is the wrong environment variable name (Medium)

The real config field is `http.addr`; `configmap.yaml` sets an env var that matches no config key at all. Masked today because the chart's default (`:8081`) happens to match the shipped `config.yaml.example`'s own default — overriding `config.httpListen` in values has no effect.

**Fix:** rename to `NEXSPENCE_HTTP_ADDR` (and, for clarity, the values key to `config.httpAddr`).

## 8. No way to configure OIDC (Medium)

`OIDCConfig` is a complete, working feature. The chart has zero `oidc` keys and no generic `extraEnv`/`extraEnvFrom` passthrough to work around the gap.

**Fix:** an `oidc:` values block mirroring `OIDCConfig`, with every field routed through `secret.yaml` (not just `client_secret`/`cookie_key` — `issuer`/`redirect_url`/`admin_group`/`groups_claim` name real internal infrastructure and access-control policy). `cookieKey` should reuse the same lookup-or-generate-and-persist pattern `config.jwtSecret` already has, so an upgrade never rotates it out from under a login in flight.

## 9. No `extraVolumes`/`extraVolumeMounts` (Medium)

`pgxpool.ParseConfig` already supports `sslmode=verify-ca`/`verify-full` + `sslrootcert=...` at the application level, but the chart's Deployment has a fixed `volumes`/`volumeMounts` list with no passthrough — there's no way to mount a CA bundle Secret/ConfigMap for a managed Postgres (RDS, Cloud SQL) that wants certificate-verified TLS, or for any other custom-mount need (LDAP/SAML over TLS, a custom root CA for outbound calls).

**Fix:** generic `extraVolumes`/`extraVolumeMounts` values (empty by default), passed straight into the Deployment's existing lists.

## 10. `NOTES.txt`'s `kubectl port-forward` command uses the wrong port (Low)

Suggests `8081:8081`, but the Service exposes port `80` by default (`8081` is only `targetPort`) — fails against the chart's own defaults.

**Fix:** `kubectl port-forward svc/{{ include "nexspence.fullname" . }} 8081:{{ .Values.service.port }}`.

## 11. Four credential values only accept plaintext, no `existingSecret` escape hatch (Medium)

Came out of testing the rest of this together: `storage.s3.accessKey`/`secretKey`, `redis.auth.password`/`externalRedis.password` (§3), `oidc.clientSecret` (§8), and `externalDatabase.dsn` (the DSN, password included) all only accept a plaintext value in `values.yaml` — no way to reference a Secret an operator already manages (an `ExternalSecret`-synced one, say), the way `postgresql.auth.existingSecret` already works for the bundled database (a real, existing Bitnami sub-chart value).

**Fix:** an `existingSecret` (Secret name) + a `...Key` (key within it) pair for each. When set, the chart's own `secret.yaml` skips composing that value, and the Deployment sources the corresponding env var straight from the named Secret via `secretKeyRef` instead — same shape as `postgresql.auth.existingSecret`. Also comment out the plaintext fields themselves by default (rather than leaving them present-but-empty next to `existingSecret`), so the Secret reference reads as the encouraged default and the plaintext alternative has to be deliberately uncommented.

jorgemillan:
Eleven fixes to the Helm chart, all independently verified with helm lint --strict / helm template / helm install --dry-run and then re-verified together with every toggle active at once.

- storage.type: s3 was non-functional - NEXSPENCE_STORAGE_DEFAULT_TYPE was never set, so the app silently stayed on local storage no matter what storage.type said. Added it to configmap.yaml.
- Liveness/readiness probes pointed at /service/rest/v1/status/check, which unconditionally returns 200 - now /healthz and /readyz.
- No way to configure Redis, so the distributed lock and login-attempt rate limiter silently degrade to no-op/per-instance once replicaCount > 1. Added redis (bundled, single-replica StatefulSet, mirroring the shape #415 gives the bundled Postgres) and externalRedis, with a redisInUse helper that fails loudly if both are enabled at once instead of silently picking one.
- No way to configure OpenTelemetry tracing, even though the app and docker-compose.yml's tracing profile both support it. Added a tracing values block mirroring TracingConfig.
- No way to set http.trusted_proxies, so every chart deployment (all of which sit behind an ingress/gateway) has the audit log and any IP-based logic see only the ingress pod's own IP. Added config.trustedProxies (comma-separated env override).
- Local storage PVC was hardcoded to ReadWriteOnce with no bring-your-own-claim option, blocking a shared-filesystem topology (e.g. EFS) for more than one replica even though LocalBlobStore's writes don't depend on POSIX locking. Added storage.local.accessMode/existingClaim.
- NEXSPENCE_HTTP_LISTEN matched no config key at all (the real one is http.addr) - renamed to NEXSPENCE_HTTP_ADDR.
- No way to configure OIDC, even though the app fully supports it. Added an oidc values block, every field routed through secret.yaml, cookieKey reusing config.jwtSecret's lookup-or-generate-and-persist pattern.
- No extraVolumes/extraVolumeMounts, so a custom CA bundle for a managed Postgres's sslmode=verify-ca (already supported at the app level) had nowhere to mount.
- NOTES.txt's kubectl port-forward command used the wrong port (8081:8081 against a Service that exposes 80 by default).
- Four credential values (storage.s3 access/secret key, redis/externalRedis password, oidc.clientSecret) plus externalDatabase.dsn only accepted plaintext, no existingSecret escape hatch the way postgresql.auth.existingSecret already has. Added the same secretKeyRef pattern to all five, and commented out the plaintext fields by default so existingSecret reads as the encouraged path.

Verified individually and all eleven combined into one chart at once, every credential sourced from an existing Secret rather than a plaintext value: helm lint --strict clean, helm template renders with no plaintext secret leaking into the manifest, helm install --dry-run confirms NOTES.txt.

Closes #422

### DIFF
--- a/deploy/helm/nexspence/templates/NOTES.txt
+++ b/deploy/helm/nexspence/templates/NOTES.txt
@@ -10,7 +10,7 @@ Nexspence has been deployed!
   Check the cilium Gateway external IP:
   kubectl get gateway {{ include "nexspence.fullname" . }}-cilium -n {{ .Release.Namespace }}
 {{- else }}
-  kubectl port-forward svc/{{ include "nexspence.fullname" . }} 8081:8081
+  kubectl port-forward svc/{{ include "nexspence.fullname" . }} 8081:{{ .Values.service.port }}
   http://localhost:8081
 {{- end }}
 
--- a/deploy/helm/nexspence/templates/_helpers.tpl
+++ b/deploy/helm/nexspence/templates/_helpers.tpl
@@ -73,3 +73,37 @@ PostgreSQL DSN — either external or bitnami sub-chart.
 {{- .Values.externalDatabase.dsn }}
 {{- end }}
 {{- end }}
+
+{{/*
+Redis mutual-exclusion guard — bundled and external are not both allowed.
+*/}}
+{{- define "nexspence.redisInUse" -}}
+{{- if and .Values.redis.enabled .Values.externalRedis.enabled }}
+{{- fail "only one of redis.enabled / externalRedis.enabled may be true" }}
+{{- end }}
+{{- end }}
+
+{{- define "nexspence.redisAddr" -}}
+{{- include "nexspence.redisInUse" . }}
+{{- if .Values.redis.enabled }}
+{{- printf "%s-redis-master:6379" .Release.Name }}
+{{- else }}
+{{- .Values.externalRedis.addr }}
+{{- end }}
+{{- end }}
+
+{{- define "nexspence.redisPassword" -}}
+{{- if .Values.redis.enabled }}
+{{- .Values.redis.auth.password }}
+{{- else }}
+{{- .Values.externalRedis.password }}
+{{- end }}
+{{- end }}
+
+{{- define "nexspence.redisDB" -}}
+{{- if .Values.redis.enabled }}
+{{- 0 }}
+{{- else }}
+{{- .Values.externalRedis.db }}
+{{- end }}
+{{- end }}
--- a/deploy/helm/nexspence/templates/configmap.yaml
+++ b/deploy/helm/nexspence/templates/configmap.yaml
@@ -6,8 +6,15 @@ metadata:
   labels:
     {{- include "nexspence.labels" . | nindent 4 }}
 data:
-  NEXSPENCE_HTTP_LISTEN: {{ .Values.config.httpListen | quote }}
+  # config.httpListen used to map to no config key at all (the real one is
+  # http.addr); NEXSPENCE_HTTP_ADDR is what viper actually reads.
+  NEXSPENCE_HTTP_ADDR: {{ .Values.config.httpAddr | quote }}
   NEXSPENCE_HTTP_BASE_URL: {{ .Values.config.baseURL | quote }}
+  {{- if .Values.config.trustedProxies }}
+  # Comma-separated; viper's default mapstructure decode hook splits a
+  # []string env override on "," (config_test.go:TestLoad_TrustedProxies_EnvOverride).
+  NEXSPENCE_HTTP_TRUSTED_PROXIES: {{ .Values.config.trustedProxies | join "," | quote }}
+  {{- end }}
   NEXSPENCE_LOG_LEVEL: {{ .Values.config.logLevel | quote }}
   NEXSPENCE_LOG_FORMAT: {{ .Values.config.logFormat | quote }}
   NEXSPENCE_AUTH_JWT_EXPIRY_HOURS: {{ .Values.config.jwtExpiryHours | quote }}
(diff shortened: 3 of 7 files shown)


skensell201:
Landed as #426 — merged into `main`, commits and authorship intact.

It conflicted with #425, which I merged a few minutes before yours and which found the `NEXSPENCE_HTTP_LISTEN` bug independently (I hit it from the other end, reviewing the Terraform provider). The resolution keeps the better half of each: your `config.httpAddr` value rename over my `httpListen` (yours matches the config key, which is the point), and your `/healthz` + `/readyz` probe paths over the old `status/check`; from #425, the `nexspence.httpPort` helper and named-port probes, retargeted at `config.httpAddr`, so `containerPort`, both probes and the Service now follow the value instead of repeating 8081.

Two things worth saying about the review:

The `NEXSPENCE_STORAGE_DEFAULT_TYPE` catch is the best one in the set. I fixed the S3 credential and `force_path_style` env names earlier today (#418) and thought that made `storage.type: s3` work — it didn't, because the app never left local storage in the first place. Half a fix is what I had.

`scripts/check-helm-env.py` (from #425) renders the chart and fails on any `NEXSPENCE_*` that matches no `SetDefault` key — it exists because that bug class hit three times today. Every environment variable your PR adds passes it: Redis, tracing, OIDC, trusted proxies, storage default type. Verified the rest live too — `/readyz` really does report its dependency checks, the redis mutual-exclusion guard fires, and all eleven toggles render together.

Closing this one as superseded. Thank you — a chart audit against what the app can actually be told is not glamorous work and this repo needed it.

Issue #404, opened 2026-09-04
Need skip login to browse open-to-all repos


kupnu4x:
There is such feature in nexus: if you don't login, you can browse public repos via web ui anyway.
It will be nice to get this behavior in nexsprence

skensell201:
Added in #420, on `main` now — you get the Nexus behaviour you described.

A visitor with no session lands on Repositories, Browse and Search instead of a login wall, and the sidebar offers a Sign in link where the user pill would be. The read endpoints behind those pages now let RBAC decide rather than demanding a token, so a repository is visible anonymously exactly when the two switches that already existed agree: instance-wide `auth.anonymous_enabled` (on by default) and the repository's own `allowAnonymous`. A repository you may not see answers 404, so its name stays unguessable. Everything else — writes, deletes, tokens, the admin pages — still requires signing in.

Going out in the next release; thank you for the suggestion.

PR #428, opened 2026-09-07
fix(ui): let a signed-out visitor actually browse public repositories (#427)

### LINKED ISSUE #427: v2.6.0: signed-out UI (#404) never renders — RepositoriesPage/BrowsePage 401s send the visitor to /login
**Summary**

On v2.6.0 the signed-out UI from #404 / #420 never renders. Opening `/` (or `/repositories`, `/browse`) without a session ends on the `/login` form on every instance, including one whose repositories all carry `allow_anonymous` with `auth.anonymous_enabled: true`.

**Where**

- `frontend/src/api/client.ts` — response interceptor: **any** `401` outside `/login` does `localStorage.removeItem('nexspence_token'); window.location.href = '/login'`, whether or not a token was ever presented.
- `frontend/src/pages/RepositoriesPage.tsx` — unconditionally queries `GET /service/rest/v1/blobstores` (store names for the row subtitle) and `GET /api/v1/repositories/:name/quota` per row. Both routes are still in the `authed` group (`router.go`), which is right — they are not visitor data.
- `frontend/src/pages/BrowsePage.tsx` — unconditionally queries `GET /api/v1/me/privileges` (and `GET /api/v1/components/:id/scan` once a docker component is selected); both `authed`.

**Reproduction (confirmed live, v2.6.0, 12 public repositories)**

Signed-out browser → `http://<host>/repositories`. Network tab:

```
GET /service/rest/v1/repositories              200   (12 public repositories — backend side of #420 works)
GET /service/rest/v1/blobstores                401
GET /api/v1/repositories/<name>/quota  ×12     401
→ window.location = /login
```

`/browse` fails the same way on `/api/v1/me/privileges`. Because the blobstores query is unconditional, the redirect fires even with zero repositories.

**Why the tests did not catch it**

`TestAnonymousBrowse_SeesOnlyPublicRepositories` exercises the router; `App.test.tsx` exercises the route table. Neither renders `RepositoriesPage`/`BrowsePage` for a guest with the live interceptor, and the page tests seed an admin session in `beforeEach`.

**Suggested fix (PR to follow)**

1. Interceptor: treat a `401` as an expired session — clear + redirect — **only when the request carried a token**. A visitor without one is left on the page; the pages already tolerate missing data.
2. `RepositoriesPage`: `enabled: signedIn` on the blobstores and quota queries.
3. `BrowsePage`: `enabled: signedIn` on privileges and scan result; render promote buttons / bulk-select checkboxes only with a session (they lead to `authed` endpoints).
4. Tests: guest render of both pages asserting no signed-in-only request is made and the location is unchanged; interceptor test for the no-token case (the existing "redirects on 401" test needs a token to keep its meaning).

Happy to take this one.

strasvet:
Closes #427.

The signed-out UI from #404 / #420 never got to render: `RepositoriesPage` queried `GET /service/rest/v1/blobstores` and `GET /api/v1/repositories/:name/quota` unconditionally, `BrowsePage` did the same with `GET /api/v1/me/privileges` (and `…/scan` per selected image), and all of those correctly stay in the `authed` group — so a visitor collected a `401` per row, and the response interceptor turned the first one into `window.location = '/login'`. On a 12-repository instance that is thirteen `401`s and a login wall; with the unconditional blobstores query it happens even on an empty instance.

## Change

- **Interceptor** (`client.ts`): a `401` is an expired session — clear the token, go to `/login` — only when the request actually carried a token. A visitor who never had one is left on the page; that is the whole premise of #404, and the pages already tolerate the missing data.
- **RepositoriesPage**: blobstores and quota queries run only with a session. Store names and usage are operator data, not visitor data.
- **BrowsePage**: the privilege query and the scan badge run only with a session; the Promote buttons and the bulk-select checkboxes (all of which lead to `authed` endpoints) are rendered only with one. Upload and delete were already gated on privileges, which a visitor has none of.
- `docs/security-rbac.md` gains a paragraph under "Browsing without signing in" saying which data the UI asks for only with a session and when a `401` means "sign in".

Nothing changes for a signed-in user: every gated query has `enabled: true` the moment a token exists, and the redirect-on-refused-token path is unchanged.

## Tests

- `client.test.ts`: the existing "redirects on 401" test now presents a token (which is what it meant); a new case sends no token, gets a `401`, and asserts the location is untouched and no `Authorization` header went out.
- `RepositoriesPage.test.tsx`: a guest render lists the repositories while the blobstores and quota handlers answer `401` and count hits — both stay at zero, the location stays put, the admin-only "New Repository" button is absent. A signed-in render asserts the same handlers are hit (blobstores once, quota once per repository).
- `BrowsePage.test.tsx`: a guest selects a repository and sees its components with the privilege handler answering `401` — zero hits, no checkbox, no Promote, no Upload, location unchanged.

Run against the unfixed code, exactly the three guest cases fail and the two signed-in cases pass. Verified in a real browser against a live instance: a signed-out visitor now gets the repository list, Browse and Search; the network tab shows no `401`.

### DIFF
--- a/docs/security-rbac.md
+++ b/docs/security-rbac.md
@@ -405,6 +405,12 @@ reads, and search. A repository the caller may not see answers `404`, not
 deletes, `/api/v1/me`, tokens, scan results, and the whole admin surface —
 still requires a token.
 
+The UI asks for that signed-in-only data (blob-store names, repository quota,
+the caller's privileges, scan results) only while a session exists, and offers
+promotion and bulk selection only then. A `401` is treated as an expired
+session — clearing it and returning to `/login` — only when the request
+carried a token; a visitor who never had one is left on the page.
+
 The Docker `/v2/` handshake itself is unconditional: an unauthenticated ping
 always answers `401` with a `Bearer` challenge pointing at `/v2/token`.
 Clients with credentials trade them there for a user token (this is the
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -106,7 +106,8 @@ describe('apiClient request interceptor', () => {
 })
 
 describe('apiClient response interceptor', () => {
-  it('redirects to /login on 401 from a non-login endpoint', async () => {
+  it('redirects to /login when a presented token is refused with 401', async () => {
+    localStorage.setItem('nexspence_token', 'expired-token')
     server.use(
       http.get('/api/v1/protected', () =>
         HttpResponse.json({ error: 'unauthorized' }, { status: 401 })
@@ -119,6 +120,21 @@ describe('apiClient response interceptor', () => {
     }
     // The interceptor sets window.location.href = '/login'
     expect(window.location.href).toBe('/login')
+  })
+
+  // A signed-out visitor browsing public repositories (#404) meets 401s from
+  // the signed-in-only endpoints; none of them is a session to lose.
+  it('leaves a signed-out visitor where they are on 401 (no token presented)', async () => {
+    let authHeader: string | null = 'unset'
+    server.use(
+      http.get('/api/v1/protected-anon', ({ request }) => {
+        authHeader = request.headers.get('Authorization')
+        return HttpResponse.json({ error: 'unauthorized' }, { status: 401 })
+      })
+    )
+    await expect(apiClient.get('/api/v1/protected-anon')).rejects.toBeDefined()
+    expect(authHeader).toBeNull()
+    expect(window.location.href).toBe('http://localhost/')
   })
 
   it('clears nexspence_token from localStorage on 401 from non-login endpoint', async () => {
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -32,13 +32,19 @@ apiClient.interceptors.request.use((config) => {
   return config
 })
 
-// Redirect to login on 401, but NOT when the 401 comes from the login endpoint itself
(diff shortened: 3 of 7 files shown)


skensell201:
Reviewed together with #430 in one worktree (both branch off `62d1bec`, they merge cleanly): `go build`/`go vet` clean, `go test ./internal/...` green apart from the sandbox-only `TestWebhookHandler_Test_200` network failure, and the frontend suite passes 545/545.

The fix reads right to me, and I checked the two things that worried me most:

- **No dead end for a visitor.** Making the interceptor stop redirecting could have left a signed-out visitor with no way to reach the login form, especially on a closed instance where the repository list answers empty rather than `401`. `Layout.tsx:357` already renders a "Sign in" link in place of the command bar when there is no session, so the affordance survives.
- **Nothing else on the guest-visible routes asks for signed-in data.** `/search` uses only `publicRead` endpoints, `/docs` fires no request at all, and `Layout`'s `system-info` query is already `enabled: admin` while the token queries live inside the profile modal a guest cannot open. The browse trees, component detail and OCI referrers the page still queries unconditionally are all on `publicRead` in `router.go`, so they answer a visitor on their own.

`ScanBadgeRow`'s `if (!signedIn) return null` sits after every hook, so the early return is safe, and gating the checkbox inside its wrapper div keeps the row grid aligned.

Thanks — clean diff, and the two new guest-render tests are exactly the coverage that was missing.

Issue #99, opened 2026-07-26
fix(group): metadata/index not merged across members; 200-on-empty shadows later members


skensell201:
## Summary
The group handler returns the **first member whose response is not 404** and never merges metadata/index documents across members (`internal/formats/group/handler.go:107`). This breaks group repositories for every format that has an aggregated index:

1. **No cross-member merge** — a group over members A and B returns only one member's index; artifacts held by the other member are invisible to clients, even though a direct artifact GET would fan out and succeed.
2. **200-on-empty shadowing (worse)** — formats whose "not found" returns HTTP **200** with an empty index (rather than 404) short-circuit fan-out at the first member, so later members are never consulted at all.

## Affected index endpoints
| Format | Endpoint | 404 on miss? |
|---|---|---|
| maven | `maven-metadata.xml` | returns member's copy, no merge |
| pypi | `/simple/<pkg>/` | **200 empty** → shadows |
| apt | `dists/.../Packages`, `Release` | **200 empty** → shadows |
| yum | `repodata/*` | **200** → shadows |
| conda | `<subdir>/repodata.json` | **200** → shadows |
| npm | packument `versions` | first member wins, no merge |
| nuget | flatcontainer version list | **200 `{"versions":[]}`** → shadows |
| gomod | `@v/list` | **200 empty** → shadows |
| terraform | `/versions` | **200 empty** → shadows |

## Reproduction (pypi example)
Group `G` = members `[hostedA, proxyB]`; package `foo` present only in `proxyB`.
```
GET /repository/G/simple/foo/
→ hostedA returns 200 with an empty "Links for foo" page
→ group returns it (code != 404, fan-out stops)
→ pip gets zero links → install of foo fails
```
(maven example: `GET /repository/G/com/foo/bar/maven-metadata.xml` returns only member A's `<versions>`, so `mvn` LATEST/RELEASE/range resolution misses B's versions.)

## Secondary: member-scoped URLs in service indexes
Several formats build absolute URLs from the member repo name, which under fan-out steers clients off the group to a single member:
- cargo `config.json` `dl`/`api` (cargo/handler.go:102-108)
- nuget service index `@id`s (nuget/handler.go:123,186)
- yum `repomd.xml` href (yum/handler.go:125,133)
- terraform `.well-known/terraform.json` (terraform/handler.go:311-317)

## Fix direction
[...]

PR #430, opened 2026-09-08
feat(cran): add R/CRAN as a new repository format

### LINKED ISSUE #429: feat(cran): add R/CRAN as a new repository format
## Summary

Add **R / CRAN** as a new repository format — hosted, proxy, and group — following the same protocol shape already implemented for APT (a generated-on-the-fly `PACKAGES` text index plus flat package files), since CRAN's own protocol is structurally the simplest of the formats not yet supported.

## Motivation

Nexspence already covers the npm/PyPI/Maven-equivalent formats for most language ecosystems, but has no repository format for R, CRAN's own language. Both Sonatype Nexus and JFrog Artifactory support CRAN, and Nexspence advertises Nexus API compatibility, so this is a conspicuous gap in the current format catalog.

CRAN's real protocol is deliberately simple:
- An index of plain-text stanzas (`src/contrib/PACKAGES`, key:value fields separated by blank lines — the same shape as Debian's `Packages` file).
- Source packages as flat tarballs, named `{package}_{version}.tar.gz`, served at `src/contrib/`.
- No signing of the index, and no meta-document describing other index documents (unlike APT's `Release`, which carries checksums of every `Packages` file it vouches for).

## Design

New package `internal/formats/cran/`, mirroring `internal/formats/apt/`'s existing pattern:

- `handler.go` — `ServeHTTP` dispatches proxy vs. hosted exactly like `apt.Handler.ServeHTTP`. `buildPackagesIndex` generates `PACKAGES`/`PACKAGES.gz` on-demand from stored assets (no index file is ever persisted, same as APT) — `Package:`/`Version:` stanzas per `.tar.gz` asset found. Upload accepts `PUT`/`POST` to a `.tar.gz` path or the repo root (normalized into `src/contrib/`, matching APT's pool-normalization for root-level uploads), plus Nexus-compatible multipart POST. Package name/version are parsed from the filename's first underscore (`cranCoords`) — simpler than APT's `debCoords`, since R package names never contain underscores, so there's no third field (architecture) to account for.
- `groupindex.go` — implements `formats.GroupIndexMerger` only, **not** `GroupIndexDependentMerger`: CRAN has no `Release`-equivalent, so there's no cross-file checksum document to keep consistent across a group's members. Stanzas are unioned across members, deduped by `Package:`+`Version:` (APT dedups by `Filename:`, which CRAN's format doesn't have).
- Wiring: new `domain.FormatCRAN`, a `repositories_format_check` migration, and a `formatRegistry["cran"]` entry in the router — no changes needed anywhere else (RBAC and the OSV.dev vulnerability-scan switch have no format-specific branches to extend; CRAN correctly falls through to the existing "vulnerability scanning is not supported for format %q" error, since OSV.dev has no CRAN ecosystem).

## Verified

Implemented and verified end-to-end against a real local stack (`docker compose up -d --build`) before writing this proposal:
- Hosted repo: uploaded a real, valid minimal R source package and installed it with a genuine `R` client (`r-base` Docker image) via `install.packages(repos=...)` — full build/install cycle succeeded, then `library()`+function call confirmed the installed package actually runs.
- Group repo: two hosted members with different packages, confirmed `PACKAGES` through the group is the real union of both (not first-member-wins), and both members' tarballs are individually downloadable through the group URL.
- Proxy repo: pointed at the real `https://cran.r-project.org/`, fetched the real (multi-MB) `PACKAGES` index unmodified, and downloaded a real package tarball (`praise_1.0.0.tar.gz`) through the proxy cache — registered with correct `name`/`version` component coordinates, cache-hit on a second request.
- Frontend: format selector, proxy-URL default (`https://cran.r-project.org/`), and badge color wired in the repositories/browse/search/cleanup pages; verified in a real Chrome session (Playwright) with no console errors.
- `go build`/`go vet`/`go test ./...` green, including new unit tests for the handler and group-merge logic mirroring APT's own test coverage.

## Out of scope (this proposal)

- **Binary packages** (Windows `.zip`, macOS `.tgz` under `bin/windows|macosx/contrib/`) — source packages only for this first version.
- **Parsing `DESCRIPTION`** out of the uploaded tarball for real `Depends:`/`Imports:`/`License:` metadata — the generated index currently derives only `Package:`/`Version:` from the filename, same minimal-metadata approach APT already takes with `.deb` filenames. `DESCRIPTION` is itself a key:value document, so it could reuse the same stanza parser as a follow-up.
- **Signing the index** — CRAN has no standard equivalent to APT's `Release`/detached-signature scheme.

**Suggested labels:** enhancement, feature, new-format

jorgemillan:
Closes #429

## Summary

Add **R / CRAN** as a new repository format — hosted, proxy, and group — following the same protocol shape already implemented for APT (a generated-on-the-fly `PACKAGES` text index plus flat package files), since CRAN's own protocol is structurally the simplest of the formats not yet supported.

## Motivation

Nexspence already covers the npm/PyPI/Maven-equivalent formats for most language ecosystems, but has no repository format for R, CRAN's own language. Both Sonatype Nexus and JFrog Artifactory support CRAN, and Nexspence advertises Nexus API compatibility, so this is a conspicuous gap in the current format catalog.

CRAN's real protocol is deliberately simple:
- An index of plain-text stanzas (`src/contrib/PACKAGES`, key:value fields separated by blank lines — the same shape as Debian's `Packages` file).
- Source packages as flat tarballs, named `{package}_{version}.tar.gz`, served at `src/contrib/`.
- No signing of the index, and no meta-document describing other index documents (unlike APT's `Release`, which carries checksums of every `Packages` file it vouches for).

## Design

New package `internal/formats/cran/`, mirroring `internal/formats/apt/`'s existing pattern:

- `handler.go` — `ServeHTTP` dispatches proxy vs. hosted exactly like `apt.Handler.ServeHTTP`. `buildPackagesIndex` generates `PACKAGES`/`PACKAGES.gz` on-demand from stored assets (no index file is ever persisted, same as APT) — `Package:`/`Version:` stanzas per `.tar.gz` asset found. Upload accepts `PUT`/`POST` to a `.tar.gz` path or the repo root (normalized into `src/contrib/`, matching APT's pool-normalization for root-level uploads), plus Nexus-compatible multipart POST. Package name/version are parsed from the filename's first underscore (`cranCoords`) — simpler than APT's `debCoords`, since R package names never contain underscores, so there's no third field (architecture) to account for.
- `groupindex.go` — implements `formats.GroupIndexMerger` only, **not** `GroupIndexDependentMerger`: CRAN has no `Release`-equivalent, so there's no cross-file checksum document to keep consistent across a group's members. Stanzas are unioned across members, deduped by `Package:`+`Version:` (APT dedups by `Filename:`, which CRAN's format doesn't have).
- Wiring: new `domain.FormatCRAN`, a `repositories_format_check` migration, and a `formatRegistry["cran"]` entry in the router — no changes needed anywhere else (RBAC and the OSV.dev vulnerability-scan switch have no format-specific branches to extend; CRAN correctly falls through to the existing "vulnerability scanning is not supported for format %q" error, since OSV.dev has no CRAN ecosystem).

## Verified

Implemented and verified end-to-end against a real local stack (`docker compose up -d --build`) before opening this PR:
- Hosted repo: uploaded a real, valid minimal R source package and installed it with a genuine `R` client (`r-base` Docker image) via `install.packages(repos=...)` — full build/install cycle succeeded, then `library()`+function call confirmed the installed package actually runs.
- Group repo: two hosted members with different packages, confirmed `PACKAGES` through the group is the real union of both (not first-member-wins), and both members' tarballs are individually downloadable through the group URL.
- Proxy repo: pointed at the real `https://cran.r-project.org/`, fetched the real (multi-MB) `PACKAGES` index unmodified, and downloaded a real package tarball (`praise_1.0.0.tar.gz`) through the proxy cache — registered with correct `name`/`version` component coordinates, cache-hit on a second request.
- Frontend: format selector, proxy-URL default (`https://cran.r-project.org/`), and badge color wired in the repositories/browse/search/cleanup pages; verified in a real Chrome session (Playwright) with no console errors.
- `go build`/`go vet`/`go test ./...` green, including new unit tests for the handler and group-merge logic mirroring APT's own test coverage.

## Out of scope (this PR)

- **Binary packages** (Windows `.zip`, macOS `.tgz` under `bin/windows|macosx/contrib/`) — source packages only for this first version.
- **Parsing `DESCRIPTION`** out of the uploaded tarball for real `Depends:`/`Imports:`/`License:` metadata — the generated index currently derives only `Package:`/`Version:` from the filename, same minimal-metadata approach APT already takes with `.deb` filenames.
- **Signing the index** — CRAN has no standard equivalent to APT's `Release`/detached-signature scheme.

### DIFF
--- a/docs/schema.sql
+++ b/docs/schema.sql
@@ -32,7 +32,8 @@ CREATE TABLE repositories (
     name            TEXT NOT NULL UNIQUE,
     format          TEXT NOT NULL CHECK (format IN (
                         'maven2', 'npm', 'docker', 'oci', 'pypi',
-                        'go', 'nuget', 'helm', 'raw', 'apt', 'yum', 'cargo', 'conan', 'conda', 'terraform'
+                        'go', 'nuget', 'helm', 'raw', 'apt', 'yum', 'cargo', 'conan', 'conda', 'terraform',
+                        'rubygems', 'cran'
                     )),
     type            TEXT NOT NULL CHECK (type IN ('hosted', 'proxy', 'group')),
     blob_store_id   UUID REFERENCES blob_stores(id),
--- a/frontend/src/pages/BrowsePage.tsx
+++ b/frontend/src/pages/BrowsePage.tsx
@@ -678,6 +678,7 @@ const FORMAT_COLORS: Record<string, string> = {
   raw: '#6b7280',
   apt: '#f59e0b',
   yum: '#10b981',
+  cran: '#276dc3',
 }
 
 // Colors for the artifact-type labels the registry browse tree reports. A media
--- a/frontend/src/pages/CleanupPage.tsx
+++ b/frontend/src/pages/CleanupPage.tsx
@@ -44,12 +44,13 @@ interface PolicyForm {
   scopePath: string
 }
 
-const FORMATS = ['*', 'maven2', 'npm', 'docker', 'oci', 'pypi', 'go', 'nuget', 'helm', 'raw', 'apt', 'yum', 'cargo', 'conan']
+const FORMATS = ['*', 'maven2', 'npm', 'docker', 'oci', 'pypi', 'go', 'nuget', 'helm', 'raw', 'apt', 'yum', 'cargo', 'conan', 'cran']
 
 const FORMAT_COLOR: Record<string, string> = {
   maven2: '#f97316', npm: '#ef4444', docker: '#3b82f6', oci: '#5b8def', pypi: '#a78bfa',
   go: '#06b6d4', nuget: '#8b5cf6', helm: '#0ea5e9', raw: '#6b7280',
   apt: '#f59e0b', yum: '#10b981', cargo: '#fb923c', conan: '#94a3b8',
+  cran: '#276dc3',
   '*': '#6b7280',
 }
 
--- a/frontend/src/pages/RepositoriesPage.tsx
+++ b/frontend/src/pages/RepositoriesPage.tsx
@@ -71,6 +71,7 @@ const FORMAT_COLORS: Record<string, string> = {
   conda:     '#44b765',
   rubygems:  '#e9573f',
   terraform: '#7b42bc',
+  cran:      '#276dc3',
 }
 
 const TYPE_LABELS: Record<string, string> = {
@@ -159,7 +160,7 @@ export default function RepositoriesPage() {
         <Select
           options={[
             { value: '', label: 'All formats' },
-            ...['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum','cargo','conan','conda','terraform','rubygems'].map(f => ({ value: f, label: f })),
+            ...['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum','cargo','conan','conda','terraform','rubygems','cran'].map(f => ({ value: f, label: f })),
           ]}
           value={formatFilter}
           onChange={setFormatFilter}
@@ -399,6 +400,7 @@ const PROXY_DEFAULTS: Record<string, string> = {
   conda:     'https://conda.anaconda.org/conda-forge/',
   rubygems:  'https://rubygems.org/',
   terraform: 'https://registry.terraform.io/',
+  cran:      'https://cran.r-project.org/',
 }
 
 /** Sets a trimmed value on cfg, or removes the key entirely when the field was cleared. */
@@ -541,7 +543,7 @@ function CreateRepoModal({ onClose, onCreated }: {
       <div className={styles.formRow}>
         <label style={LABEL_STYLE}>Format</label>
         <Select
-          options={['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum','cargo','conan','conda','terraform','rubygems'].map(f => ({ value: f, label: f }))}
+          options={['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum','cargo','conan','conda','terraform','rubygems','cran'].map(f => ({ value: f, label: f }))}
           value={form.format}
           onChange={handleFormatChange}
         />
--- a/frontend/src/pages/SearchPage.tsx
+++ b/frontend/src/pages/SearchPage.tsx
@@ -94,6 +94,7 @@ const S = {
 const FORMAT_COLORS: Record<string, string> = {
   maven2: '#f97316', npm: '#ef4444', docker: '#3b82f6', oci: '#5b8def', pypi: '#a78bfa',
   go: '#06b6d4', nuget: '#8b5cf6', helm: '#0ea5e9', raw: '#6b7280', apt: '#f59e0b', yum: '#10b981',
+  cran: '#276dc3',
 }
 
 function fmtSize(b: number | undefined) {
@@ -296,7 +297,7 @@ export default function SearchPage() {
             <Select
               options={[
                 { value: '', label: 'any' },
-                ...['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum'].map(f => ({ value: f, label: f })),
+                ...['maven2','npm','docker','oci','pypi','go','nuget','helm','raw','apt','yum','cran'].map(f => ({ value: f, label: f })),
               ]}
               value={filters.format}
               onChange={v => setFilters(f => ({ ...f, format: v }))}
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -29,6 +29,7 @@ import (
 	"github.com/nexspence-oss/nexspence/internal/formats/cargo"
 	"github.com/nexspence-oss/nexspence/internal/formats/conan"
 	"github.com/nexspence-oss/nexspence/internal/formats/conda"
+	"github.com/nexspence-oss/nexspence/internal/formats/cran"
 	"github.com/nexspence-oss/nexspence/internal/formats/gomod"
 	"github.com/nexspence-oss/nexspence/internal/formats/group"
 	"github.com/nexspence-oss/nexspence/internal/formats/helm"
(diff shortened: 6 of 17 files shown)


skensell201:
Reviewed together with #428 in one worktree (both branch off `62d1bec`, they merge cleanly): `go build`/`go vet` clean, `go test ./internal/...` green apart from the sandbox-only `TestWebhookHandler_Test_200` network failure, and the frontend suite passes 545/545. The handler mirrors `apt` closely, which makes it easy to follow, and the real-shape proxy test plus the group-merge tests are the right coverage.

One functional gap before this goes in, on the group surface:

**A group with a proxy member serves the upstream index instead of the merged one, because R asks for `PACKAGES.rds` first.**

`GroupIndexSourcePath` claims `/PACKAGES` and `/PACKAGES.gz`, but not `PACKAGES.rds`. So a request for it falls through to the first-non-404 fan-out in `group/handler.go` — a hosted member answers `405` (the `default:` case), but a **proxy** member answers `200` with `cran.r-project.org`'s own `PACKAGES.rds`, and that is what the group returns.

That matters because `PACKAGES.rds` is not a fallback for R, it is the preferred document. From `utils::available.packages` ([r-source `src/library/utils/R/packages.R:123`](https://github.com/wch/r-source/blob/trunk/src/library/utils/R/packages.R#L123)): it downloads `PACKAGES.rds` and `readRDS`es it first, and only if *that* download or read raises an error does it try `PACKAGES.gz`, then plain `PACKAGES`.

So on the usual `group(cran-hosted, cran-proxy)` layout — the same shape the npm/PyPI groups are deployed in — an R client resolves against the upstream index alone and never sees a single hosted package, while `install.packages` still reports success for whatever it does find. That is exactly the shadowing #99 introduced the merger to prevent; the group-merge tests do not catch it because they only exercise the two paths `GroupIndexSourcePath` claims.

A hosted-only group is safe today only by accident: every member answers `405`, R errors and falls back to the merged `PACKAGES.gz`.

The minimal fix is to stop the group from answering `PACKAGES.rds` at all, so R falls back to the merged index it can read — e.g. claim the path in `GroupIndexSourcePath` and have `MergeGroupIndex` refuse it, rather than letting fan-out relay a member's copy. Generating a real RDS is not worth it (it would mean writing R's serialization format in Go); a document R cannot read is enough, since its own fallback chain then does the right thing. It would also be worth having the hosted handler answer `404` rather than `405` for `PACKAGES.rds`, since a GET of an index this repository does not keep is a "not found", not a bad method — and a group of hosted members would then merge rather than relay.

Two smaller notes, neither blocking:

- `buildPackagesIndex` lists every `.tar.gz` asset in the repository regardless of whether it sits under `/src/contrib/`, so a tarball uploaded to some other path (raw-style) would appear in the index at a URL R would then fail to fetch. `apt` has the same shape, so this is consistent — just worth a `strings.HasPrefix(a.Path, "/src/contrib/")` guard while it is cheap.
- The `Limit: 1000` on both the component and asset lookups is copied from `apt`; same known cap, same eventual pagination work. Fine to leave as is here.

Thanks for the thorough write-up in #429 and for verifying against a real `R` client — the end-to-end evidence made this much quicker to review.

jorgemillan:
Thanks for the thorough review — the PACKAGES.rds gap was real, and the reference to utils::available.packages's actual download order made it straightforward to track down.

Pushed in 19bae25:

GroupIndexSourcePath now claims PACKAGES.rds onto the same plain PACKAGES source .gz already uses.
MergeGroupIndex refuses to produce one — there's no way to merge stanzas into a binary RDS — which degrades to a member's plain-text body. That's not valid RDS either, but readRDS/available.packages's own fallback treats a read error just like a failed download and retries PACKAGES.gz, the flavor that actually merges.
Also made the hosted handler answer 404 instead of 405 for a GET it doesn't recognize, so a group of all-hosted members can fan out past it to the next one too.
Verified end-to-end against a real docker compose stack: a group with a real proxy member (cran.r-project.org) and a hosted-only package, hit by a real R client. The access log confirms R requests PACKAGES.rds first, gets the new non-RDS response, and retries PACKAGES.gz — which resolves and installs the hosted-only package the old fan-out would have hidden behind the proxy's real index. Also confirmed the package renders correctly in Browse over a real Chrome session.

Left the two smaller notes (the /src/contrib/ prefix guard, the hardcoded Limit: 1000) as-is since you flagged them non-blocking — happy to fold them into this PR too if you'd rather have them here than as a follow-up.

skensell201:
Re-reviewed `19bae25` on top of #428 in the same worktree: `go build` clean, the cran and group packages pass 91/91, and the new `TestGroupMerge_CranNeverRelaysProxyRDS` pins the exact failure mode (proxy listed **first**, which is what actually beat the old fan-out).

The shadowing is gone — I probed the three paths directly against a group of `proxy(cran-upstream), hosted`:

```
GET /repository/pg/src/contrib/PACKAGES.rds  → 200 application/octet-stream  "Package: upstreamonly\nVersion: 1.0\n"
GET /repository/p-hosted/src/contrib/PACKAGES.rds → 404
GET /repository/pg/src/contrib/PACKAGES      → 200 text/plain  (merged union)
```

The upstream `.rds` never reaches the client, and R's `readRDS` fails on that body exactly as your end-to-end run showed, so it falls back to the merged `.gz`. The hosted `404`-instead-of-`405` is right too, and I like that the comment says why the path has to be claimed even though it cannot be answered.

**One consequence of the degrade path worth a follow-up, not a blocker:** refusing the merge lands in `serveMergedIndex`'s non-strict branch, which serves `parts[0].Body` — the **first member's whole plain `PACKAGES`**. In the test that is 36 bytes; against a real `cran.r-project.org` proxy member it is the upstream index, on the order of 15–20 MB for ~22k packages. So every `available.packages()` against a group whose first member is a CRAN proxy transfers that index to the client as a bogus `.rds`, throws it away, and then fetches the merged `.gz` — twice the bytes, for a document that is never usable.

Two ways out, both small:

- Implement `formats.GroupIndexStrictMerger` on the cran handler with `GroupIndexMemberFailureIsFatal` returning `false` for every path. `strict` is only consulted through that callback in `collectIndexParts`, so member-skipping stays exactly as it is today, while the merge refusal takes the `isStrict` branch — a short `502` instead of a member's document. R treats a failed download the same as a failed `readRDS`, so the fallback chain is unchanged.
- Or return an empty body with a nil error for `.rds`, which is cheap but keeps the `200`.

I'd take the first: a `200` whose body is a different document is the part that could surprise a caching intermediary, and the interface already exists for "this could not be produced".

The two earlier notes are fine as follow-ups — no need to grow this PR. Nice work tracking it down from the client's own source.

Issue #438, opened 2026-09-09
README's missing two formats and still calls tracing "planned"


jorgemillan:
Noticed the README's fallen a bit behind in a couple of spots — nothing urgent, just flagging it.

**Two formats missing from the table.** `internal/formats/` has 16 formats implemented but the "Supported Package Formats" table only lists 15. RubyGems is missing, and so is R/CRAN (just merged in #430) — it didn't get added as part of that PR.

**Roadmap marks OpenTelemetry tracing as "planned".** It's fully implemented already, since #302 (`internal/tracing/tracing.go`, wired into `cmd/server/main.go`, its own config section, tests and all). Tracing landed on August 22nd and the README hasn't been touched since August 18th, so that row just never got updated. Of the two things marked "planned", only SBOM generation/cosign signing is genuinely still pending.

While I'm here — any interest in adding more stuff to that Roadmap? A couple of ideas, no pressure either way:
- Alpine (apk) — the cheapest one to add: same index+package pattern that already exists for the APT/Yum handlers, so it reuses almost all the existing plumbing.
- Hugging Face model repositories — much more expensive, since its real protocol runs over git+git-lfs (or its REST API with large files served via CDN), nothing like the index+package formats that exist now. But neither Nexus nor JFrog covers this well yet, so it might be worth the extra work.

skensell201:
Thanks — both were right, and the format table was further behind than the count suggested: the Group column still said "—" for APT, Yum, Conan, Conda and Terraform, all of which merge group indexes today. Fixed in #444, along with a guard test (`internal/domain/readme_formats_test.go`) holding the README's count and table to `domain.AllFormats`, the same way the docs site is already held to it — so the next format that lands fails CI instead of drifting.

On the Roadmap ideas: Alpine is in review as #440 (thank you for that one too). Hugging Face is a fair call about the gap in the market, but as you say its protocol shares nothing with the index+package formats here — no plans yet, happy to keep discussing it.

Will ship in the next release.

PR #440, opened 2026-09-09
feat(alpine): add Alpine (apk) as a new repository format

### LINKED ISSUE #439: feat: Alpine (apk) as a new repository format
## Summary

Nexspence supports two other OS-package formats (APT, Yum/RPM) but not Alpine's `apk`. This proposes adding it, following the same shape as APT: an on-the-fly generated index (no persisted index file), hosted + proxy + group repository types, coordinates derived from the uploaded package.

## Protocol

Alpine resolves a repository URL by appending `/<arch>/` itself (no `dist`/`component` split like APT):

```
GET  /<arch>/APKINDEX.tar.gz     → index (generated on the fly, gzip-wrapped tar with a single "APKINDEX" entry)
GET  /<arch>/<file>.apk          → package download
PUT  /<arch>/<file>.apk          → upload (hosted; a repository convention, not part of the apk wire protocol itself — same status as APT/Yum's own upload endpoints)
DELETE /<arch>/<file>.apk        → delete
```

`APKINDEX` is a third index serialization Nexspence doesn't have yet: one-letter-key stanzas (`P:`/`V:`/`A:`/`C:`/`S:`/`I:`/`T:`/`L:`/`D:`/`p:`, blank-line separated) — closer in shape to APT's `Packages` than to Yum's XML, just a different vocabulary.

## Two things worth flagging up front (found by testing against real packages from dl-cdn.alpinelinux.org, not obvious from the format spec alone)

1. **The `C:` checksum is not "hash the whole file."** An `.apk` is a concatenation of independent gzip members: an *optional* signature wrapper (one per signing key — every officially published package carries one), then the control tar.gz (`.PKGINFO` + install scripts), then the data tar.gz. `C:` is `"Q1" + base64(SHA1(control tar.gz))` — the control member specifically, whichever gzip member is the first one that is *not* a signature wrapper. Assuming "the control segment is always the first member" breaks the moment a real signed package (i.e. anything mirrored from an official CDN) is re-uploaded — hashing the signature instead produces a checksum a real apk client's own recomputation never matches, and it rejects the package as corrupt. Detecting a signature member just means checking whether the member's tar first entry is named `.SIGN.<algo>...`.
2. **The index needs `.PKGINFO` parsed for `D:` (depends) to be useful at all.** Without it, `apk add <pkg>` "succeeds" (the package itself downloads and verifies fine) but the binary then fails at runtime with relocation/missing-symbol errors, because the client was never told it needs the shared libraries the package links against. `.PKGINFO` is a flat `key = value` text file (not stanza-shaped like `APKINDEX`) with repeatable `depend =` / `provides =` lines; it's also the authoritative source for `pkgname`/`pkgver`/`pkgdesc`/`license`/`size` (installed/uncompressed size) — the real `apk index` tool derives the index entirely from it, not from the uploaded filename.

## Signing

Real Alpine repos are typically signed via `abuild-sign` (an RSA signature block over the unsigned `APKINDEX.tar.gz`, appended as a leading tar entry named `.SIGN.RSA.<key>.pub` — a different mechanism from APT's OpenPGP clearsign/detached signatures, not reusable from that code). Out of scope for a first version: an unsigned index is a real, supported client mode (`apk add --allow-untrusted`), the same posture Yum already has (dnf doesn't require repo signing by default either).

## Suggested design

- New package mirroring `apt`'s structure: index built per-request from `Components.Search`/`Assets.List` (filtered by `.apk` suffix and by the `<arch>` path segment), wrapped in a single-entry `tar.gz` on serve.
- `GroupIndexMerger` only (no `GroupIndexDependentMerger`-equivalent — unlike APT's `Release`/Yum's `repomd.xml`, there's no root document describing other documents to keep consistent across a group's members).
- Hosted upload: parse `.PKGINFO` from the control segment for name/version/description/license/depends/provides/installed-size, falling back to the `<name>-<version>-r<rel>.apk` filename convention only if `.PKGINFO` is missing or unparseable.

jorgemillan:
Same shape as APT: on-the-fly generated APKINDEX.tar.gz, hosted + proxy + group, coordinates and dependency metadata (D:/p:/T:/L:/I:) derived from the uploaded package's real .PKGINFO.

Two things verified against real packages from dl-cdn.alpinelinux.org: the C: checksum must be computed over the first gzip member that is NOT a signature wrapper (a real published .apk is signed, so the control segment is not always the first member), and the index needs .PKGINFO parsed for D: to be useful at all — without it apk installs a package without rejecting it, but the binary fails at runtime because the client was never told which shared libraries it needs.

Closes #439


skensell201 (changes requested):
Thanks — this is a strong first version, and the two things you flagged up front are exactly the two that are easy to get wrong: computing `C:` over the first non-signature gzip member (rather than the first member) and parsing `.PKGINFO` for `D:` are both right, and `controlSegment`'s `countingReader`/`Multistream(false)` pair is the correct way to find a member boundary in Go. The migration, the `AllFormats` entry, the group merger and the real-shape proxy test all follow the shape the last few formats set.

One blocker and three smaller things.

## 1. `pathArch` reads the first path segment, not the architecture directory (blocker)

`pathArch` takes the segment after the leading `/`, which is only the architecture when the repository root is the directory that directly contains `<arch>/`. Alpine's own published layout is `<branch>/<repo>/<arch>/`, and a hosted repository mirroring it is an ordinary thing to want:

```
/x86_64/APKINDEX.tar.gz              -> arch="x86_64"   ok
/v3.20/main/x86_64/APKINDEX.tar.gz   -> arch="v3.20"    wrong
/edge/community/aarch64/APKINDEX.tar.gz -> arch="edge"  wrong
```

(that's the output of `pathArch` on this branch, not a guess).

The failure is silent, which is what makes it worth fixing before merge rather than after: `buildIndex` then filters assets on `HasPrefix(a.Path, "/v3.20/")` — which matches every architecture under that branch — and writes `A:v3.20` into every stanza. The index is served `200` with a full-looking body, and a real `apk` on x86_64 discards every entry because the arch field is not its own. `apk add` reports the package as not found, with nothing in the server log to suggest why.

The architecture is the directory the requested file sits in, in every layout including the single-level one:

```go
dir := path.Dir(p)             // "/v3.20/main/x86_64"
arch := path.Base(dir)         // "x86_64"
prefix := dir + "/"            // filter assets on this instead of "/"+arch+"/"
```

That also makes the asset filter exact rather than prefix-of-a-branch, so a group of packages under a sibling architecture can no longer leak into another architecture's index. Worth a test case per row of the table above.

## 2. Hitting `maxMemberBytes` produces a wrong checksum instead of an error

`decompressOneGzipMember` reads through `io.LimitReader(gz, maxMemberBytes)`. When a member decompresses to more than 8 MiB, `ReadAll` stops at the limit without an error, the gzip trailer is never read, and `cr.n` therefore counts fewer source bytes than the member actually occupies. `controlSegment` then returns `apk[offset : offset+consumed]` — a truncated compressed range — and `checksumQ1` hashes it happily. The upload is accepted, the index publishes a `C:` no client can reproduce, and the package fails verification at install time with nothing pointing back at the upload.

8 MiB is a generous bound for a real control segment, so this is unlikely to fire on a genuine `.apk` — but the failure mode should be an error, not a wrong answer. Reading one byte past the limit and erroring if it is there is enough.

## 3. A signed `APKINDEX.tar.gz` is untested

`mergeIndexParts` skips a member whose index it cannot unpack (`continue`), which is the right call for a malformed member but means a break here silently drops that member's entire package list from the group — the exact shadowing the merger exists to prevent, just one level down.

Every index a proxy member serves from a real mirror is signed: `gzip(tar with .SIGN.RSA.<key>.pub, no end-of-archive blocks)` concatenated with `gzip(tar with APKINDEX)`. I built that shape against this branch and `unpackIndexTarGz` handles it correctly — the entry loop, rather than "take the first tar entry", is what saves it. Since nothing in the suite covers it, please add it as a fixture (a hand-built two-member `.tar.gz`, no real key needed) so it stays covered.

## 4. Small things

- `Limit: 1000` on both lookups in `buildIndex` is copied from `apt`/`cran`, so a repository past 1000 packages serves a short index. Not yours to fix here — it's now #443, covering all three formats — but worth a comment pointing at it, the way the cap is not currently acknowledged.
- The README's format table is not updated. It's already missing two formats (#438); please leave it to that issue rather than growing this PR.
- `parsePKGInfo` matches the entry name exactly as `.PKGINFO`. abuild writes it that way, so this is fine — mentioning it only because a `./`-prefixed tar entry is a common variation and a one-line `strings.TrimPrefix` would make it moot.

Happy to merge once (1) is fixed; (2) and (3) are small enough to fold into the same push.

jorgemillan:
Thanks for the thorough review — all three points fixed in `ffefca8`, plus the `#443` pointer.

## 1. `pathArch` (blocker) — fixed

`pathArch` now takes `path.Base(path.Dir(p))` instead of the first path segment, and `buildIndex` filters assets by the exact directory (`path.Dir(p)+"/"`) instead of `"/"+arch+"/"`. Added the three cases from your table to `index_test.go`, plus a multi-level fixture in `handler_test.go` (`TestAlpine_Index_FiltersByArch_MultiLevelLayout`).

Verified live against the exact scenario you flagged: published a real `curl-8.14.1-r2.apk` under `/v3.20/main/x86_64/` and a real `lua-lunix-20190403-r0.apk` under `/v3.20/community/aarch64/` — same leading `v3.20` segment, different arch. A real `apk` client in `alpine:3.20` installed `lua-lunix` cleanly from the aarch64 index, and the x86_64 index kept showing only `curl` (checksum matched Alpine's real `APKINDEX` byte for byte). Before this fix both packages would have collided into one index.

## 2. `maxMemberBytes` silent truncation — fixed

`decompressOneGzipMember` now reads `maxMemberBytes+1` and returns an explicit error when a member hits that ceiling, instead of silently truncating and hashing a partial segment. Added a test that forces a >8MiB decompressed member and asserts the error.

## 3. Signed `APKINDEX.tar.gz` — covered

Added a two-member fixture (`.SIGN.RSA.*` member + real `APKINDEX` member) to `groupindex_test.go`, confirming `mergeIndexParts`/`unpackIndexTarGz` correctly pulls the package list out of the second member.

## Small things

- Added `// see #443` next to both `Limit: 1000` lookups in `buildIndex`.
- Left the README table and `.PKGINFO` prefix handling untouched, per your notes.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./internal/formats/alpine/... -v` (32/32) and the full `go test ./...` all green.
- Live end-to-end in `docker compose`: real hosted repo, real `.apk`s from `dl-cdn.alpinelinux.org`, real `apk` client install, plus a Chromium/Playwright pass over `/browse` confirming both packages render correctly split by architecture with no console errors.

skensell201 (approved):
All three review points are addressed — `pathArch` now takes the last directory component and `buildIndex` filters by the exact directory (with the multi-level layout test), the oversized gzip member is rejected instead of silently truncated, and the signed-index fixture covers the `.SIGN.RSA.*` member. Verified together with the other open PRs in one tree: build, vet, golangci-lint and the full test suite are green. Thanks!

PR #448, opened 2026-09-11
feat(auth): let local users change their own password and admins reset any local account

n00bi14:
## What

A local user can change their own password from **avatar → Profile → Change
Password**, and an admin can reset any local account's password from **Admin →
Security → Users**. Both paths enforce `auth.password_min_length`, refuse
identity-provider accounts, and end the affected user's browser sessions.

## Behaviour

| Route | Current password | Who |
|---|---|---|
| `PUT /api/v1/me/change-password` | **always required** | any authenticated local user, admins included |
| `PUT /service/rest/v1/security/users/{userId}/change-password` | not required | `nx-admin` only |

- **`auth.password_min_length` is enforced for the first time.** The setting has
  existed in config for a long time and was never read by anything. It now gates
  every password write — user creation, self-service change, admin reset — in the
  service layer, so a direct API call cannot slip past it. A minimum of 0 means
  "not wired" and disables the check, which keeps existing fixtures working.
- **ldap / oidc / saml accounts are refused (403).** The login flow never compares
  against the stored hash for them, so writing a local hash used to "succeed" and
  do nothing. The refusal lives in the service, not just in the UI; the Users page
  disables the reset action for every non-local source rather than only for ldap.
- **A change revokes the user's JWTs** by bumping `tokens_valid_after`, including
  the caller's own session — the profile modal signs them out and says so. `nxs_*`
  API tokens are separate credentials and keep working.
- The admin branch used to answer 500 for every service error. A refused reset is
  the caller's fault: 403 for an identity-provider account, 400 for a password
  below the minimum.

## Review fixes in this branch

A review of the first eight commits turned up one real defect and a few things
the UI claimed but did not deliver:

- **`ChangePassword` picked its verb from the caller's role, not the route.** An
  `nx-admin` hitting `/api/v1/me/change-password` fell into the admin branch and
  ran `SetPassword`, which takes no old password at all — so the profile modal
  asked for the current password and the server discarded it. A stolen session, a
  leaked API token or an unattended browser could take over the account (the
  default bootstrap `admin` above all) without knowing the password it replaced.
  The absent `:userId` is what identifies the self route, so it branches on that
  now. Covered by handler tests and an integration test; the admin route keeps its
  no-old-password reset, including on the admin's own account.
- **The admin-reset half of the feature had no end-to-end coverage.**
  `TestAdminResetsLocalUserPassword` now walks it: old password stops working, new
  one logs in.
- **`auth.password_min_length` was only discoverable through a 400.**
  `/api/v1/auth/config` serves the value, and the three forms that write a password
  mirror it as a hint plus a pre-flight check. The service layer stays the
  authority — an unknown minimum simply submits.
- **The profile modal overstated what it revoked** ("every existing session") and
  the Users page said nothing when an admin reset their *own* password, leaving the
  next request to bounce to `/login` unexplained. Both now say what actually
  happens.

## Docs

- `docs/api-spec.yaml` described the admin route's body as `text/plain` and did not
  mention `/api/v1/me/change-password` at all. Both routes are specified now,
  including which one verifies the current password and how identity-provider
  accounts are answered.
- The website's self-service section described the behaviour the role-based
  branching produced ("any non-admin user can rotate their own password") and its
  curl example sent `current_password` / `new_password`, which the handler does not
  bind — the documented call would have answered 400. Fixed in EN and RU.
- `docs/deployment.md` documents the setting; `docs/architecture.md` records the
  route-not-role rule so the next change does not reintroduce it.

## Testing

`go build` clean · `golangci-lint` 0 issues · backend unit tests green ·
`tsc --noEmit` and `eslint` clean · frontend 564/564 · integration tests
(`-tags=integration`, Docker + Postgres) green, including the three password
scenarios.

`TestWebhookHandler_Test_200` fails in my environment because netguard blocks
loopback — it fails identically on `main` and is unrelated to this branch.

## Not in scope

- `must_reset_password` is set by the Nexus migration (which assigns a random
  password nobody knows) but is never enforced or surfaced anywhere, so those
  accounts cannot get in. A forced-change flow is its own change.
- `/api/v1/me/change-password` is not throttled; `loginguard` only covers login.
  It needs a valid session, but it is still a free oracle for the current password.


skensell201 (approved):
Reviewed on top of the current main (after #445/#447) and built together with the other open PRs: build, vet, lint, Go and frontend suites are green. The self-route fix is the important part — an `nx-admin` on `/api/v1/me/change-password` really did land in `SetPassword` with the current password discarded. Enforcing `auth.password_min_length` (default 8) in the service layer is the right place; it also means `POST /users` with a shorter password now answers 400, which is what the setting always promised. Thanks!

PR #450, opened 2026-09-11
feat(huggingface): add Hugging Face Hub as a new repository format

### LINKED ISSUE #449: feat: Hugging Face Hub as a new repository format — first approximation
## Summary

Add Hugging Face Hub (models, datasets, spaces) as a new repository format — hosted, proxy, and group, matching the pattern already used for Terraform Registry and CRAN. Both huggingface_hub (1.31.0) and `transformers`/`diffusers` work against it unmodified by pointing `HF_ENDPOINT` at a repository.

## Why

Hugging Face is a growing gap in the artifact-manager space: neither this project nor the other self-hosted registries have mature support for it yet, while it's become one of the primary distribution channels for ML models and datasets, right alongside PyPI/npm/Maven for their own ecosystems.

The protocol turns out to be simpler than it looks. The real backend uses git+git-lfs (and, increasingly, a newer content-addressable backend called Xet — see below) internally, but a client **never speaks either protocol to download**: `huggingface.co` resolves everything server-side before answering a plain `GET .../resolve/{revision}/{filename}`. That's what makes hosted + proxy + group tractable without reimplementing git.

## Protocol surface

```
GET  /api/{models|datasets|spaces}/{repo_id}                       → repo metadata (siblings, sha)
GET  /api/{models|datasets|spaces}/{repo_id}/revision/{revision}   → metadata at a revision
GET  /api/{models|datasets|spaces}/{repo_id}/tree/{revision}[/dir] → file tree
GET  /{repo_id}/resolve/{revision}/{filename}                      → download (models; datasets/spaces get a /datasets//spaces prefix)
HEAD /{repo_id}/resolve/{revision}/{filename}                      → file metadata (ETag, commit, size)
PUT  /{repo_id}/resolve/{revision}/{filename}                      → publish (hosted only — see Non-goals)
```

`repo_id` is `namespace/name`, or a bare `name` for a canonical repo (`bert-base-uncased`). Since it can be one or two segments, it can't be split by counting path segments — the handler locates it by the literal `/resolve/`, `/tree/`, or `/revision/` marker instead, the same trick the existing Terraform handler already uses for `/upload/`/`/download/`.

## What's implemented

- **Hosted**: metadata/tree computed from stored assets; publish via a plain per-file `PUT` at the same URL the file downloads from (a Nexspence convention, not the real Commit API — see Non-goals). A revision's reported commit sha is derived from the revision's own contents (path + sha256 of every file), not just its name, so re-publishing under the same branch changes the sha a client sees.
- **Proxy**: binary downloads are cached like any other proxied artifact. A commit sha is immutable and cached forever; a branch/tag is revalidated on the metadata TTL (default 10 min, configurable per repo) like any mutable index. Metadata/tree calls are relayed 1:1 — unlike Terraform's provider JSON, a Hub metadata document names files relatively (`siblings[].rfilename`), so there's no absolute URL to rewrite.
- **Group**: metadata and tree are merged across members instead of shadowed by the first one that answers (the same class of bug #99 fixed for apt/yum/CRAN groups).

## Real-world verification, and what it found

This was implemented and then verified against the real `huggingface_hub` client (1.31.0) and real `huggingface.co` — not just unit tests with a mocked upstream — because three real protocol behaviors turned out to be invisible from a mock:

1. **`HEAD` on an LFS/Xet-backed file loses its own metadata if the client transparently follows the redirect.** A real Hub host answers `HEAD` on a large file with a redirect to its CDN, and `X-Repo-Commit`/`X-Linked-ETag`/`X-Linked-Size` live *only* on that redirect response — the CDN's own final response carries none of them. A client that only sees the final hop refuses to download at all (`FileMetadataError`). `huggingface_hub`'s own redirect follower already handles this (it stops at the first redirect that leaves the host and reads metadata off *that* response); the proxy handler now does the same manually instead of letting its HTTP client auto-follow.

2. **`X-Xet-Hash` sends the client into a protocol this format doesn't implement.** Most of `huggingface.co`'s real weight files today (`model.safetensors`, `pytorch_model.bin`, `tf_model.h5` — not the small text files) are stored on Xet, a newer content-addressable backend. The client treats that one header as the trigger to fetch through a three-party CAS protocol (a `xet-read-token` call back to the Hub, then chunk reconstruction against a third host) instead of a plain download. Relaying the header verbatim sent every real client into that path and then into a 404. The fix withholds it: the client falls back to the plain HTTP path it already offers for non-Xet-aware clients, and `huggingface.co` keeps a fallback URL behind the same redirect specifically for that — bytes confirmed identical by checksum either way.

3. **A hard, whole-request timeout silently caps every proxied download at a fixed duration, independent of progress.** This isn't specific to this format — it's in the shared proxy HTTP client used by all proxy-capable formats — but Hugging Face is what surfaced it: a real 6.4 GB model checkpoint (`gpt2-xl/model.safetensors`) proxied against `huggingface.co` was cut off mid-transfer at a flat 5-minute mark regardless of whether the transfer was progressing normally. See the dedicated section below.

Verified end to end after the fixes: a real model (`bert-base-uncased`), a real ~416 MB dataset shard (`legacy-datasets/wikipedia`), and a real space (`gradio/hello_world`) all proxy with checksums matching a direct download; a real 6.4 GB model checkpoint (`gpt2-xl/model.safetensors`) downloads in one pass with a checksum matching a direct download; a 3 GB hosted publish+download round-trips correctly while the server's memory stays flat (true streaming, not buffered); proxy TTL revalidation was confirmed with real timing (a mutable branch reference picks up an upstream change only after the configured TTL elapses; an immutable commit-sha reference never does, matching git's own semantics).

## A shared-infrastructure fix, and the alternatives considered

Finding #3 above lives in the shared proxy HTTP client (`http.Client.Timeout`), not in the Hugging Face handler — it's used identically by every proxy-capable format (Docker, npm, PyPI, Conda, Terraform, and the rest). Three ways to handle it were on the table:

- **Fix it in the shared client** (chosen). Replace the whole-request `Timeout` with `Transport.ResponseHeaderTimeout` (bounds connect + wait-for-headers only, unchanged at 5 minutes) plus a small idle-read wrapper on the response body that fails a single `Read` after 2 minutes with no bytes. This catches a hung or silently-stalled connection exactly as before, without capping how long a large, still-progressing transfer may take — which is the only difference that matters for a cache whose whole job is mirroring artifacts of whatever size an upstream happens to publish. Chosen because a format-local workaround would leave the same bug open for every other proxy-capable format, and because the fix is small, mechanical, and directly testable (own unit tests for the idle-read wrapper, plus the real 6.4 GB download as a live check).
- **Format-local workaround.** A separate HTTP client with no timeout, used only for Hugging Face's own binary downloads, leaving the shared client untouched. Faster to land and lower-risk in isolation, but leaves the identical bug in place for every other format's large artifacts (a big Conda package, a large Docker layer, a big npm/PyPI wheel) — rejected as treating a symptom instead of the cause.
- **Leave it and document it.** Lowest risk, but the bug is real and reproducible today for any proxy-capable format's large-enough artifact, not hypothetical.

The fix was re-verified against every other proxy-capable format after the change (their existing real-upstream-shaped test suite, plus a live `pip install` and a live `npm install` through their own proxy repos with checksums matching direct downloads) to confirm nothing regressed from touching shared code.

## Non-goals (v1)

- **The real Hub Commit API** (preupload + LFS-batch-like negotiation + atomic git commit) is not implemented — it's backend-specific and unverifiable without huggingface.co's own infrastructure. Hosted publish uses a plain per-file `PUT`, the same convention Terraform and raw already use for their own hosted uploads, documented as a Nexspence convention rather than the real protocol.
- **The Xet CAS protocol itself** is not implemented (see finding #2) — proxying degrades a Xet-backed file to the classic HTTP path rather than reproducing Xet's own chunk-level dedup transport. Same bytes, no content-defined deduplication on the wire.
- **Synthetic `oid`/blob ids**: tree entries carry a synthetic oid (the asset's own sha256, truncated) rather than a real git blob id, since there's no git object graph behind a hosted repository. `RepoFile` reads that field unconditionally, so it has to be present; it's never claimed to be a real git object id.
- No real Hugging Face authentication (`HF_TOKEN`) semantics beyond the existing generic JWT/API-token auth already used by every other format.

jorgemillan:
Closes #449.

## Summary

Adds Hugging Face Hub (models, datasets, spaces) as a new repository format — hosted, proxy, and group, following the same pattern as the existing Terraform Registry and CRAN formats. Both `huggingface_hub` (1.31.0) and `transformers`/`diffusers` work against it unmodified by pointing `HF_ENDPOINT` at a repository — no client-side changes needed.

## Protocol surface

```
GET  /api/{models|datasets|spaces}/{repo_id}                       → repo metadata (siblings, sha)
GET  /api/{models|datasets|spaces}/{repo_id}/revision/{revision}   → metadata at a revision
GET  /api/{models|datasets|spaces}/{repo_id}/tree/{revision}[/dir] → file tree
GET  /{repo_id}/resolve/{revision}/{filename}                      → download (models; datasets/spaces get a /datasets//spaces prefix)
HEAD /{repo_id}/resolve/{revision}/{filename}                      → file metadata (ETag, commit, size)
PUT  /{repo_id}/resolve/{revision}/{filename}                      → publish (hosted only — see Non-goals)
```

`repo_id` is `namespace/name`, or a bare `name` for a canonical repo (`bert-base-uncased`). Since it's one or two segments, it can't be split by counting path segments — the handler locates it by the literal `/resolve/`, `/tree/`, or `/revision/` marker instead, the same trick the existing Terraform handler already uses for `/upload/`/`/download/`.

The real backend uses git+git-lfs (and, increasingly, a newer content-addressable backend called Xet) internally, but a client never speaks either protocol to download: `huggingface.co` resolves everything server-side before answering a plain `GET .../resolve/{revision}/{filename}`. That's what makes hosted + proxy + group tractable without reimplementing git.

## What's implemented

- **Hosted**: metadata/tree computed from stored assets; publish via a plain per-file `PUT` at the same URL the file downloads from (a convention, not the real Commit API — see Non-goals). A revision's reported commit sha is derived from the revision's own contents (path + sha256 of every file), not just its name, so re-publishing under the same branch changes the sha a client sees.
- **Proxy**: binary downloads are cached like any other proxied artifact. A commit sha is immutable and cached forever; a branch/tag is revalidated on the metadata TTL (default 10 min, configurable per repo) like any mutable index. Metadata/tree calls are relayed 1:1 — unlike Terraform's provider JSON, a Hub metadata document names files relatively (`siblings[].rfilename`), so there's no absolute URL to rewrite.
- **Group**: metadata and tree are merged across members instead of shadowed by the first one that answers (the same class of bug #99 fixed for apt/yum/CRAN groups).

## Real-world verification, and what it found

Implemented and verified against the real `huggingface_hub` client (1.31.0) and real `huggingface.co`, not just unit tests with a mocked upstream — six real bugs across two verification passes turned out to be invisible from a mock, since it never actually redirects or streams multiple gigabytes:

1. **`hf_hub_download` refuses to download without both `ETag` and `X-Repo-Commit`** on the `resolve/` response — confirmed load-bearing, not cosmetic. A hosted revision's commit is derived from that revision's contents so a re-publish is actually reflected, not just its branch name.
2. **`snapshot_download` asks the tree endpoint at a commit sha**, not just `resolve/` — a sha→branch mapping present only on the download path failed the snapshot on its second call. All three metadata endpoints now do the mapping.
3. **Publish `Content-Type` can't come from the client**: `curl -X PUT --data-binary` labels its body `application/x-www-form-urlencoded`, which was being stored and served back on every download. The file extension decides instead, with `.safetensors`/`.onnx`/`.parquet`/`.bin` (unknown to Go's mime table) falling back to `application/octet-stream`, matching what `huggingface.co` itself serves them as.
4. **`HEAD` on an LFS/Xet-backed file lost its own metadata** because the HTTP client auto-followed the redirect to the CDN — `X-Repo-Commit`/`X-Linked-ETag`/`X-Linked-Size` live *only* on that redirect response, confirmed against `bert-base-uncased`'s real `pytorch_model.bin`. Fixed by manually replicating `huggingface_hub`'s own redirect follower (stop at the first cross-host hop, read metadata off that response, answer 200 so the client's follow-up `GET` still comes back through the proxy).
5. **`X-Xet-Hash` sent the client into a protocol this format doesn't implement.** Most of `huggingface.co`'s real weight files today (`model.safetensors`, `pytorch_model.bin`, `tf_model.h5`) are stored on Xet, a newer content-addressable backend; that header alone triggers a three-party CAS fetch this proxy doesn't speak, landing in a 404. Withholding it makes the client fall back to the plain HTTP path it already offers non-Xet-aware clients — same bytes, confirmed by checksum.
6. **A hard, whole-request 5-minute timeout in the shared proxy HTTP client** (`internal/formats/repoproxy`, used by all proxy-capable formats, not specific to this one) silently capped every proxied download regardless of progress — a real 6.4 GB model checkpoint (`gpt2-xl/model.safetensors`) was cut off mid-transfer. Replaced `Client.Timeout` with `Transport.ResponseHeaderTimeout` (connect + wait-for-headers only, unchanged at 5 min) plus a new `idleReader` that fails a `Read` after 2 minutes with no bytes — catches a stalled connection without capping a large, still-progressing one. The same file then downloaded in one clean pass (244s, 26.4 MB/s, no retries) versus 1382s and several outright failures before the fix, with an identical checksum to a direct download either way. Re-verified against every other proxy-capable format (their existing real-upstream-shaped tests, plus a live `pip install` and `npm install` through their own proxies with checksums matching direct downloads) to confirm nothing regressed from touching shared code.

Also verified live: a real model (`bert-base-uncased`), a real ~416 MB dataset shard (`legacy-datasets/wikipedia`), and a real space (`gradio/hello_world`) all proxy with checksums matching direct downloads; a 3 GB hosted publish+download round-trips correctly with server memory staying flat throughout (true streaming, never buffered); proxy TTL revalidation confirmed with precise real timing (a mutable branch picks up an upstream change only after the TTL elapses, an immutable commit sha never does).

## Non-goals (v1)

- **The real Hub Commit API** (preupload + LFS-batch-like negotiation + atomic git commit) isn't implemented — backend-specific and unverifiable without huggingface.co's own infrastructure. Hosted publish uses the same plain per-file `PUT` convention Terraform and raw already use.
- **The Xet CAS protocol itself** isn't implemented (see finding #5) — a proxied Xet-backed file degrades to the classic HTTP path rather than reproducing Xet's own chunk-level dedup transport. Same bytes, no content-defined deduplication on the wire.
- **Synthetic `oid`/blob ids**: tree entries carry the asset's own sha256 (truncated) rather than a real git blob id, since there's no git object graph behind a hosted repository. `RepoFile` reads that field unconditionally, so it has to be present.
- No Hugging Face-specific authentication (`HF_TOKEN`) beyond the existing generic JWT/API-token auth every other format already uses.

## Testing

`go build`/`go vet`/`go test ./...` and the `-tags integration` suite are green, as are the 553 frontend tests. New coverage: `internal/formats/huggingface/*_test.go` (handler, proxy including two redirect-hop scenarios and the Xet-header-drop case, group-index merge), `internal/formats/group/merge_huggingface_test.go`, a proxy real-shape case in `internal/integration/proxy_formats_test.go`, and `internal/formats/repoproxy/idlereader_test.go` for the new shared idle-timeout wrapper.


skensell201 (changes requested):
Thanks — the format itself reads well, and the live-client findings (ETag/X-Repo-Commit on HEAD, the cross-host redirect hop, dropping `X-Xet-Hash`) are exactly the kind of thing a mocked upstream never shows. I built this branch together with #440 and #462 in one tree: build, vet, golangci-lint, the Go suite and the 572 frontend tests are green on the union.

Two things before merge, one of them about the shared `repoproxy` change.

## 1. The idle-timeout guard covers only `repoproxy`'s own copy sites

Removing `Client.Timeout` is the right call — a fixed 5-minute cap on a multi-gigabyte body was wrong. But the replacement is applied per call site (`idleReader{resp.Body, …}` inside `repoproxy.go`, plus `FetchUpstreamOnce` wrapping its own response), while the client itself is shared: **terraform, conda, nuget and helm** call `repoproxy.ClientFor(repo).Do(req)` directly and read the body themselves (`internal/formats/{terraform,conda,nuget,helm}/handler.go`), and this PR adds two more such sites (`serveProxyAPI` / `upstreamRequest` here). Before this change those reads were bounded by the 5-minute `Client.Timeout`; after it they have no bound at all on a stalled body — the handler sits until the client goes away.

The fix that closes this in one place is a `RoundTripper` wrapper on the transport (both `UpstreamClient` and `buildProxyClient`) that wraps `resp.Body` with the idle guard for every response. Then every caller of `ClientFor(repo)` gets the stall bound automatically, the per-site wrapping in `repoproxy.go` and the `idleReadCloser` in `FetchUpstreamOnce` can go, and nobody has to remember to wrap in the next format.

## 2. One goroutine + timer per `Read`

`idleReader.Read` spawns a goroutine and a `time.After` for every call. `io.Copy` reads in 32 KiB chunks, so the 6.4 GB checkpoint from the description is ~200k goroutine spawns; and after a timeout fires the abandoned `Read` still holds the caller's buffer. A single watchdog per body is cheaper and has no leftover goroutine writing into `p`: record `time.Now()` atomically on every successful `Read`, and one goroutine per body (started on first `Read`, stopped by `Close`) that closes the body when `now - lastProgress > idle`. The net/http body then returns an error from the blocked `Read` on its own, which is the same thing the current design relies on.

## Small things

- After #440 merged this conflicts in the format registries (`docs/schema.sql`, `router.go`, the four `*.tsx` lists, `DocsPage.tsx`, `types.go`, the two site/README guard tests) — all "both added" — and the migration needs renumbering: `032` is now taken by alpine, and #462 also carries a `032`, so please take `033` (or `034` if #462 lands first). The README/site format count guard tests will want 19.
- For a proxy repository every `HEAD …/resolve/…` goes upstream, so an HF proxy cannot serve a cached model while huggingface.co is unreachable — fine for v1, but worth one line in the docs section, since "the cache still works offline" is what people assume a proxy repository gives them.

jorgemillan:
Thanks for the thorough review — both blockers made sense, and the shared-client angle on #1 was exactly right. Pushed `f774cdb` addressing both, plus the small items.

## 1. Idle guard moved to the transport

Replaced the per-call-site `idleReader` wrapping with a `RoundTripper` (`idleGuardedTransport`) wrapped around `UpstreamClient`'s transport and every client `buildProxyClient` builds. Every response from either client now gets its body wrapped automatically, regardless of which handler reads it — `repoproxy.go`'s own copy sites, terraform/conda/nuget/helm's direct `ClientFor(repo).Do(req)` reads, and this PR's `serveProxyAPI`/`upstreamRequest`/`upstreamHeadHop` are all covered the same way, with nothing for the next format to remember to wrap.

Verified live against real upstreams through the refactored client (not just unit tests): `registry.terraform.io` (versions JSON via the exact `ClientFor().Do()` + direct-body-read pattern terraform/conda/nuget/helm use), `api.nuget.org` (service index, same pattern), and `huggingface.co` itself — HEAD on a real LFS-backed file, the metadata API (`serveProxyAPI`, the previously-unguarded call site), and a real GET download — all unchanged in behavior, all through a live `docker compose` stack, cross-checked in Chrome with zero console errors.

Added `TestIdleGuardedTransport_ProtectsDirectBodyReads` as a regression test: a raw `client.Do(req)` + unwrapped `io.ReadAll(resp.Body)` (the terraform/conda/nuget/helm shape) against a handler that answers headers and then stalls — confirms the body errors out on the idle timeout even with no wrapping at the call site.

## 2. Single watchdog per body

Replaced `idleReader`'s per-Read goroutine+timer with `idleTimeoutBody`: one watchdog goroutine per response body, started when the transport wraps it and stopped by `Close`. It tracks `lastProgress` (atomic, updated on every successful `Read`) and force-closes the underlying body once `idle` has elapsed since the last one — which is what makes the currently-blocked `Read` return an error on its own, same mechanism as before, just triggered by the watchdog instead of left to the caller's own `Close`.

`TestIdleTimeoutBody_SingleWatchdogAcrossManyReads` checks the goroutine count stays flat across ~500 reads and drops back down after `Close` — no more one-goroutine-per-`Read`.

## Small things

- Rebased onto `main` now that #440 (Alpine) merged; resolved the "both added" conflicts in the format registries. Migration renumbered `032` → `033` (Alpine took `032`) — the `Up` constraint now lists both `alpine` and `huggingface` so an Alpine repo doesn't get dropped from the CHECK when this migration runs after it; `Down` reverts to the alpine-only list.
- Documented the offline-cache caveat in `DocsPage.tsx`: `huggingface_hub` always sends a HEAD before downloading and that HEAD is never answered from cache, so `huggingface.co` being unreachable still fails `hf_hub_download`/`snapshot_download` even for an already-cached file. (Narrower than "every resolve/" — a direct GET by immutable commit sha skips upstream entirely, and a GET by branch/tag only revalidates once its cached copy is older than `metadata_max_age`.)

`go build`/`go vet`/`go test ./...` green (37 packages, plus `-tags integration` build), and the full live-compose + Chrome pass above. golangci-lint isn't available in the environment I verified in, so that one I can't confirm directly — happy to run it if you point me at how you invoke it in CI, or if it flags something I'll fix it as a follow-up.

skensell201 (approved):
Both blockers from the 15.09 review are closed the way I hoped:

1. idleGuardedTransport at the transport level means every client from buildProxyClient and UpstreamClient itself guards the body regardless of call site — terraform/conda/nuget/helm's direct `ClientFor(repo).Do` reads are now covered without touching them, and TestIdleGuardedTransport_ProtectsDirectBodyReads pins exactly that shape. Dropping Client.Timeout in favour of ResponseHeaderTimeout + idle watchdog is the right split for a streaming proxy; the gpt2-xl cut-off is a convincing reproduction.
2. idleTimeoutBody: one watchdog goroutine per body, atomic lastProgress, stopped by Close; the goroutine-count test guards against the per-Read regression.

Checked that nothing type-asserts the client transport to *http.Transport (only DefaultTransport in storage/s3 and azure, untouched), so the wrapper cannot break a caller.

Housekeeping I pushed to your branch (maintainerCanModify) instead of another round-trip: merged main and renumbered the migration 033 → 036 — 033 went to the Azure blob store (#462) the day you rebased, 034 to #493 and 035 to #495. Merging when CI is green.

PR #462, opened 2026-09-14
feat(storage): add Azure Blob Storage as a new backend

n00bi14:
## Summary

Introduces Azure Block Blob Storage as a full third storage backend alongside local storage and S3 — including configuration, admin UI, Helm deployment, and documentation.

- **Storage backend:** New Azure backend with streaming uploads, chunked appends over staged blocks (O(bytes) instead of O(n²) for OCI pushes), and presigned URLs (via account key or a cached user-delegation key). Supports connection string, account key, SAS token, or an ambient Entra ID identity as credential sources. Lifecycle rules are rejected outright, since Azure scopes them to the whole storage account rather than to individual containers.
- **Configuration:** `storage.default_type` is now validated (a typo like `"S3"` previously fell through silently to the local backend). Seed stores (default/docker) are only repointed to the configured container on startup while they are still untouched and empty — otherwise startup fails loudly instead of leaving asset references pointing at an empty container.
- **Fail closed on store errors:** An assigned but unreachable blob store used to silently fall back to the default store (writes landed in the wrong store, reads/deletes came back empty). Fixed across all call sites: artifact store, fetch/delete, proxy cache, docker browse, npm age filter, and the OCI mount path.
- **Garbage collection:** References are now also indexed by physical store identity (endpoint + container), so GC no longer deletes blobs still referenced through a second logical store pointing at the same container (e.g. the seeded default/docker stores sharing one container).
- **API & admin UI:** Azure stores can now be configured from the admin UI. All four credential types are redacted and answered with a per-key marker, so an edit that omits a credential keeps the stored value while an explicit empty value clears it. More than one Azure credential at once is rejected instead of silently resolved by precedence.
- **Helm/deployment:** Azure added as a third storage type in the Helm values, including a workload-identity example for AKS (no stored credentials needed). An unknown `storage.type` now aborts the chart render instead of producing a release that silently doesn't do what the values say. Documents that switching backends on an installation whose seed stores already hold blobs refuses to start, along with the migration path via the blob-store migration endpoint.
- **Misc:** MinIO test image is now pulled from Quay instead of Docker Hub (Docker Hub no longer serves `minio/minio`); fixed an SSRF-guard issue blocking the live webhook delivery test.

## Migration

Includes migration `032_blob_store_azure.sql` (widens the blob store type constraint). Its `Down` step is deliberately `NOT VALID` — existing `azure` rows are left alone on rollback rather than being aborted or pointed at a filesystem path they never had.


skensell201:
Read through the whole thing and built it together with #440/#450 in one tree — build, vet, golangci-lint, the Go suite and the frontend tests are green on the union. The fail-closed store resolution is the right behaviour (an unreachable named store silently falling back to the default was a data-placement bug waiting to happen), the physical-identity index in GC only ever *adds* protection, and thanks for catching `/usage` returning the store unredacted — that was leaking the S3 `secret_key` to any admin reading usage.

One thing before merge: **the migration number.** `032_blob_store_azure.sql` collides with `032_format_alpine.sql`, which landed with #440 a few minutes ago (and #450 carries a `032` too). goose refuses duplicate versions, so please rebase onto main and renumber to `033`. main also just took #448, which touches `cmd/server/main.go` and `router.go` — expect a trivial conflict there.

Once that's rebased I'm happy to approve; the rest reads well.

n00bi14:
I will change it in a few hours after work.

skensell201 (approved):
Rebased cleanly, the migration is `033` now and CI is green on top of v2.7.0 — approving. Thanks for the thorough work, and for the `/usage` redaction catch along the way.

PR #467, opened 2026-09-16
feat(helm): existing Istio Gateway, existing Secrets, and Trivy DB mirrors

fallen-up:
## Summary
- Attach the VirtualService to a Gateway CR that already exists (`gateway.istio.existingGateway`); skip creating one. `gatewaySelector` is only the workload labels on a chart-created Gateway (`spec.selector`), not a Gateway name.
- Read JWT and bootstrap admin (username + password) from a Secret that already exists, same pattern as DSN / S3 / OIDC client secret.
- Pass Trivy CVE DB mirrors via `scanning.dbRepository` / `javaDbRepository` (lists, matching `scan.trivy.db_repository`).

## Test plan
- [x] `helm lint deploy/helm/nexspence`
- [x] `python3 scripts/check-helm-env.py`
- [x] `helm template` with `gateway.istio.existingGateway` — VirtualService points at that Gateway, no Gateway CR
- [x] `helm template` with `jwtSecretExistingSecret` + `adminExistingSecret` — those keys are omitted from the chart ConfigMap/Secret and injected via `secretKeyRef`
- [x] `helm template` with `scanning.enabled` + DB lists — `NEXSPENCE_SCAN_TRIVY_DB_REPOSITORY` / `JAVA_DB_REPOSITORY` are comma-joined


skensell201 (approved):
Rendered the chart with each new value set locally — an existing Gateway (namespaced `ns/name` included) replaces the chart's own, the JWT/admin keys leave the chart's ConfigMap/Secret when an existing Secret is named and arrive via `secretKeyRef`, and the Trivy DB lists comma-join into the env names `scan.trivy.*_db_repository` already read. Guard script and lint are green, and you extended both for the new cases. Thanks for a tidy first contribution!

PR #468, opened 2026-09-16
build(deps-dev): move to vitest 5 with its coverage provider, and group them for dependabot

skensell201:
Supersedes #456. `@vitest/coverage-v8` peer-depends on `vitest` at the exact same version, so Dependabot's lone bump of the coverage provider to 5.0.0 failed `npm ci` with ERESOLVE — the same trap the react and typescript-eslint packages already had.

- `vitest` 4.1.11 → 5.0.0 and `@vitest/coverage-v8` 4.1.11 → 5.0.0 together. No config or test changes needed: 32 files / 564 tests pass and the coverage thresholds hold (83.8% statements).
- `.github/dependabot.yml`: a `vitest` group (`vitest`, `@vitest/*`) so the two keep moving as one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skensell201:
CI caught what the local run did not: the `typecheck:test` step. vitest 5 dropped the `jest` namespace mirror that jest-dom's default types entry augments, so every `toBeInTheDocument()` lost its type. Switched `setup.ts` and `tsconfig.test.json` to `@testing-library/jest-dom/vitest`, which augments vitest's own `Assertion` — typecheck, build, tests and lint pass locally.

PR #478, opened 2026-09-16
fix(config): keep dotted hostnames in docker subdomain aliases

### LINKED ISSUE #469: Docker subdomain aliases with dotted hostnames fail to parse (Viper key delimiter)
## Summary

`docker.subdomain_connector.aliases` is `map[string]string` and the docs show FQDN keys:

```yaml
aliases:
  docker-hub-proxy.domain.example.com: dockerhub-proxy
```

`config.Load` unmarshals via Viper with the default key delimiter `.`. After YAML parse, those hostnames are split into nested maps, and Unmarshal fails:

```
parse config: decoding failed due to the following error(s):
'docker.subdomain_connector.aliases[docker-hub-proxy]' expected type 'string', got unconvertible type 'map[string]interface {}'
```

Quoting the YAML key does not help — the split is after parse.

## Where

`internal/config/config.go` (`v.Unmarshal`) + `docker.subdomain_connector.aliases`

The Helm chart mounts a small `config.yaml` for aliases (`nexspence-file`); the rendered YAML is valid. This is not a chart bug.

## Reproduction

`config.yaml` (or the chart-mounted overlay):

```yaml
docker:
  subdomain_connector:
    enabled: true
    aliases:
      "docker-hub-proxy.example.com": "dockerhub-proxy"
```

Confirmed on 2.7.0. Changing the **value** to a name without dots does not help; the **key** is what Viper splits (`aliases[docker-hub-proxy]`).

## Impact

Hostname → repository aliases cannot be used with real DNS names. The feature added for Nexus-style hostnames (#282) does not load.

## Suggested fix

Keep Viper's `.` delimiter for nested keys (`http.addr`, …). Flatten nested maps back into dotted keys when unmarshalling into `map[string]string`.

fallen-up:
Fixes #469

## Summary
Viper's default key delimiter is `.`, so FQDN keys in `docker.subdomain_connector.aliases` (e.g. `docker-hub-proxy.example.com`) are split into nested maps. Unmarshal into `map[string]string` then fails with `aliases[docker-hub-proxy] expected string, got map`. Quoting the YAML key does not help.

Flatten those nested maps back into dotted hostnames on decode. Keep Viper's duration and comma-slice hooks (passing `DecodeHook` replaces them).

## Test plan
- [x] `go test ./internal/config/ -count=1`
- [x] Config with quoted and unquoted FQDN alias keys loads; hostnames stay intact


skensell201 (approved):
Checked the one thing that made me pause: passing `DecodeHook` replaces viper's defaults, and viper 1.21 actually ships an unexported `stringToWeakSliceHookFunc` (any slice kind) rather than `StringToSliceHookFunc` (exactly `[]string`). Every slice in `Config` is `[]string`, so the two are equivalent here — and the comma-joined env lists (`scan.trivy.*_db_repository`, the Helm feature you just added) still decode. The hook also quietly fixes `role_mappings` for group names with dots (Google-style `team.dev@…`), which #481/#483 are circling around. Nice catch on "quoting doesn't help — the split is after parse".

PR #487, opened 2026-09-16
fix(auth): don't wipe manually-assigned roles when OIDC/SAML sends no group info

### LINKED ISSUE #481: OIDC role sync silently wipes manually-assigned roles when the id_token has no groups claim at all
## Summary

`LoginOIDC` calls `syncOIDCRoles` unconditionally on every login. `syncOIDCRoles` builds its wanted-role set purely from `claims.Groups`, then always calls `SetUserRoles` with whatever it found — REPLACE semantics, by design, so the IdP is the source of truth. But when the id_token simply carries no groups claim under any name (Google Workspace without Admin SDK/Directory API — no OIDC scope adds group membership to a standard id_token; this is nexspence's own documented setup for Google, see `docs/oidc-setup.md` line ~72), `claims.Groups` comes back empty for a structurally different reason: not "the IdP says this user is in zero groups" but "the IdP never told us anything about groups." The code doesn't distinguish the two, so it treats "no info" the same as "confirmed zero groups" and wipes the role set — including any role an admin granted manually via Security > Users — on every single login. `docs/oidc-setup.md`'s own troubleshooting table (line ~177) tells operators to "set `groups_claim: ""` and assign roles manually" for Google, but the code never actually honored that promise.

The same unconditional-REPLACE-with-no-presence-check pattern also exists in `syncLDAPRoles` (LDAP) and the SAML equivalent — not touched by this fix, flagged here for awareness since the same fix shape would apply.

## Where

- `internal/service/user_service.go` — `LoginOIDC` (call site), `syncOIDCRoles`
- `internal/auth/oidc.go` — `OIDCClaims`, `extractClaims`

## Impact

Any OIDC deployment where the IdP doesn't or can't supply group claims has manual role assignment silently undone on the affected user's very next login, with no error and no log warning distinguishing it from a real "IdP says zero groups" case — indistinguishable from a random access outage to the admin who just granted it. Reported from a real deployment (Google Workspace SSO): an admin assigns a role, the user logs out and back in, the role is gone.

## Suggested fix

Distinguish "the id_token has no claim under `GroupsClaim` at all" from "the claim is present but an empty list" — only the latter is real signal from the IdP and should trigger REPLACE; the former means "this IdP gives no group info for this login," so leave the user's existing role assignments untouched.

Add a `GroupsPresent bool` to `auth.OIDCClaims`, set in `extractClaims` via a plain key-presence check (`_, ok := raw[s.cfg.GroupsClaim]`) alongside the existing `getStrSlice` extraction. `syncOIDCRoles` takes this flag and returns immediately (no `SetUserRoles` call) when it's false.

\`\`\`go
_, groupsPresent := raw[s.cfg.GroupsClaim]
return &OIDCClaims{
    ...
    Groups:        getStrSlice(s.cfg.GroupsClaim),
    GroupsPresent: groupsPresent,
    ...
}
\`\`\`

\`\`\`go
func (s *UserService) syncOIDCRoles(ctx context.Context, userID string, groups []string, groupsPresent bool) error {
    if !groupsPresent {
        return nil
    }
    ...
}
\`\`\`

**Verified 2026-09-16:** implemented as above; `go build`/`go vet`/`go test ./...` green (37 packages), including new unit coverage (`TestLoginOIDC_NoGroupsClaim_PreservesManualRoles`, `TestLoginOIDC_GroupsClaimPresentButEmpty_StillReplaces` in `internal/service`; `TestOIDCService_NoGroupsClaimInToken_GroupsPresentFalse` in `internal/auth`) confirming both the fix and that real REPLACE semantics are untouched for IdPs that do send groups (a present-but-empty claim still wipes roles, as intended). Also verified live end-to-end against a real `docker compose` stack (real Postgres, real Keycloak as the OIDC IdP) driven by a real headless Chrome session (Playwright): configured `groups_claim: ""` and `admin_group: ""` — matching a real Google-Workspace-shaped deployment — logged a Keycloak test user into nexspence via a real OIDC redirect round-trip (JIT-provisioned, zero roles), had an admin assign a role via the real Users API, then logged the same user out and back in through a second real OIDC round-trip. Before the fix: the reissued JWT's `roles` claim came back `[]` (confirmed both in the token and in the `oidc login complete` server log line). After the fix, same exact steps: the manually-assigned role survives the second login, in both the token and the server log.

jorgemillan:
LoginOIDC/LoginSAML sync roles with REPLACE semantics on every login, which is correct when the IdP actually reports group membership. But when the id_token/assertion carries no groups claim/attribute at all (e.g. Google Workspace OIDC without Admin SDK — no scope adds group membership to a standard id_token), the code couldn't distinguish that from "IdP confirmed zero groups", and wiped every manually-assigned role on the user's very next login with no error or warning.

Add GroupsPresent to OIDCClaims/SAMLClaims, set via a plain key/attribute presence check in extractClaims for both. syncOIDCRoles/syncSAMLRoles now no-op when it's false, leaving existing role assignments untouched, and log a warning when groups_claim/groups_attribute is configured (non-empty) but still absent from the token — signal of a real misconfiguration, as opposed to the intentional groups_claim: "" case, which stays silent.

Verified live end-to-end against a real docker-compose stack (real Postgres, real Keycloak): with NEXSPENCE_OIDC_GROUPS_CLAIM="", a role granted manually via the Users API survived a second OIDC login round-trip (previously wiped to []). Reconfiguring to a real but wrong claim name reproduced the new warning log while still preserving the role. SAML has no live IdP in the dev compose stack; its coverage is unit tests against real crewjam/saml Assertion/AttributeStatement structs (the same types production parsing consumes).

The equivalent LDAP bug (a group *search* that can fail or be unconfigured, rather than a *claim* that can be absent) has a different root cause and is filed separately.

Closes #481


skensell201 (approved):
Checked together with #488's LDAP counterpart (#500) and the rest of today's batch in one worktree over main — build, vet, lint, unit + integration all green. The presence check in extractClaims is the right place: it is the only spot that still sees the raw token/assertion, so the service layer gets a single boolean instead of guessing from an empty slice. Merging.

PR #490, opened 2026-09-16
feat(backup): configurable scheduled backup; fix Export/Restore ignoring per-asset blob store

### LINKED ISSUE #489: Backup export/restore ignores per-asset blob store, plus a configurable scheduled backup
## Summary

Two related findings while adding a scheduled-backup option on top of the existing manual Backup & Restore:

1. **Bug:** `BackupService.Export`/`Restore`/`ImportRepo` always read/wrote blobs through the instance's single default `BlobStore`, ignoring each asset's actual `BlobStoreID`. Every other blob-touching service (`CleanupService`, `GCService`, `ReplicationService`, blob-store migration) already resolves each asset's real physical store via `StoreResolver` — backup was the one exception.
2. **Feature:** a configurable scheduled backup (cron + destination blob store + retention), reusing the same `distlock`/`robfig-cron` pattern those services already use for HA-safe scheduling.

### Where
- `internal/service/backup_export.go` — `writeBlobEntries` (`s.BlobStore.Get(ctx, a.BlobKey)`, ignoring `a.BlobStoreID`)
- `internal/service/backup_import.go` — `restoreAssets`/`importRepoAssets` (`s.BlobStore.Put(...)`, same issue on the write side)

### Impact
A repository pinned to a non-default blob store (S3, Azure — anything other than the instance's configured default) silently loses its assets from `Export`: `writeBlobEntries` treats a failed `Get` against the wrong store as "unreadable, skip" — no error, no warning, the archive is just missing those blobs. `Restore`/`ImportRepo` has the mirror problem: it writes the bytes to the default store while the asset row still names the original (correct) store, so after a restore the asset's metadata and its actual bytes point at two different places.

### Suggested fix
Add a `Resolver StoreResolver` to `BackupService` (same interface `CleanupService`/`GCService` already use) and a `storeFor(ctx, blobStoreID)` helper that resolves the real store per asset, falling back to the default only when resolution genuinely isn't possible (no resolver wired, or the store lookup fails) — matching `CleanupService.storeForAsset`'s existing fallback semantics. Used in `writeBlobEntries` (keyed off `a.BlobStoreID`) and both restore paths (keyed off the resolved/remapped destination store ID for that asset).

**Verified live:** created a repository backed by a dedicated MinIO (S3) bucket, uploaded a real file, ran a manual `Export`, deleted the repository, restored from the archive through the real UI (Restore button, driven by a real Chrome session) — the asset came back with the same `blobStoreId`, byte-identical content, and the blob physically landed back in the same S3 bucket (confirmed via `mc ls` against MinIO directly, and confirmed the instance's local default store received nothing).

### Design

A singleton `backup_settings` row (`enabled`, `scheduleCron`, `blobStoreId`, `retentionCount`), editable from Admin → Backup & Restore without a restart:

```yaml
enabled: true
scheduleCron: "0 3 * * *"
blobStoreId: "<uuid of a blob store created via Admin → Blob Stores>"
retentionCount: 7   # 0 = unlimited
```

`BackupService.RunScheduled`:
- Exports to a temp file first (`Put` needs a known size; the archive's final size isn't known until `Export` finishes streaming it — the same spool-then-upload approach `backup_import.go` already uses for the inbound side), then `Put`s it to the resolved destination under `backups/nexspence-backup-<timestamp>.tar.gz`.
- **Resolution failure for the chosen destination is a hard error, not a fallback to the default store** — unlike the per-asset case above, silently falling back here would defeat the entire point of choosing a separate destination and hide exactly the kind of failure this feature exists to catch. (Caught this distinction live: an early version reused the lenient per-asset fallback, and a misconfigured destination store silently sent several scheduled backups to local disk instead of the configured MinIO bucket — undetected until the destination bucket was inspected directly.)
- Retention deletes the oldest `backups/` entries beyond `retentionCount`, via `BlobStore.ListEntries` (already exposes `ModTime`, no new storage-interface surface needed).
- Guarded by the same `distlock` pattern `CleanupService`/`GCService`/`ReplicationService` already use, so a multi-replica deployment runs the scheduled backup once per tick, not once per replica.
- Each run's outcome (success/failure) is also written to Security → Audit Log (`domain: SYSTEM`, `action: BACKUP`) — no existing cron job does this; `AuditMiddleware` only ever sees real HTTP requests, so a cron-triggered run needed its own direct `AuditRepo.Write`.

New endpoints: `GET`/`PUT /api/v1/backup/settings`.

**Verified live end-to-end:** real `docker compose` stack (Postgres, MinIO), a real 1-minute cron schedule during verification driving 3 consecutive real runs — retention correctly kept only the 2 most recent, confirmed via `mc ls` against MinIO directly (not mocked) — settings saved and reloaded through the real UI (Chrome), including a full page reload confirming server-side persistence, and a real Security → Audit Log entry for a successful run.

## Out of scope
- No incremental backup — each scheduled run is a full snapshot, same as manual export.
- No automatic restore — restoring a scheduled backup is the same manual Restore button, by design (an automatic remote restore would be a destructive action nobody asked for).
- No notification (email/webhook) on success/failure — structured logging + the audit log entry only, matching this codebase's other cron jobs.

jorgemillan:
Closes #489

## Summary

Two things bundled together, since the second was found while building the first:

1. **Fix:** `BackupService.Export`/`Restore`/`ImportRepo` always read/wrote blobs through the instance's single default `BlobStore`, ignoring each asset's actual `BlobStoreID`. Every other blob-touching service (`CleanupService`, `GCService`, `ReplicationService`, blob-store migration) already resolves per-asset via `StoreResolver` — backup was the one exception. A repo on a non-default S3/Azure store silently lost its blobs from `Export` and got them written to the wrong store on `Restore`/`ImportRepo`.
2. **Feature:** a configurable scheduled backup — cron schedule, destination blob store, retention — on top of the existing manual Export/Restore.

## What changed

- `BackupService` gains a `Resolver StoreResolver` + `storeFor` helper, used by `writeBlobEntries` (Export/ExportRepo) and both restore paths, so each asset's blob is read/written through its real physical store instead of always the default.
- New singleton `backup_settings` table (migration `034`), editable from Admin → Backup & Restore (`GET`/`PUT /api/v1/backup/settings`) without a restart.
- `BackupService.RunScheduled` exports to a temp file (`Put` needs a known size, unlike the streaming HTTP export path) and writes it to the resolved destination. Resolution failure here is a hard error, never a silent fallback to the default store — that fallback is appropriate for the per-asset case above, but for the destination an admin explicitly chose, silently falling back to the default would hide exactly the kind of failure this feature exists to catch.
- Retention deletes the oldest `backups/` entries beyond `retentionCount` via `BlobStore.ListEntries` (already exposes `ModTime` — no new storage-interface surface).
- Guarded by the same `distlock` pattern `CleanupService`/`GCService`/`ReplicationService` already use, so a multi-replica deployment runs the scheduled backup once per tick, not once per replica.
- Each run's outcome is also written to Security → Audit Log (`domain: SYSTEM`, `action: BACKUP`) — no existing cron job does this; `AuditMiddleware` only sees real HTTP requests, so a cron-triggered run needed its own direct `AuditRepo.Write`.
- New "Scheduled Backup" card in Admin → Backup & Restore: enable toggle, cron input, destination dropdown (from existing blob stores), retention input, last-run status.

## Test plan

- `go build`/`go vet`/`go test ./...` green, including a new Postgres integration test for `BackupSettingsRepo` and a regression test asserting the cross-store fix (`TestBackup_ExportRestore_NonDefaultBlobStore`) and the no-silent-fallback behavior (`TestBackupService_RunScheduled_ResolutionFailure_ReturnsErrorNotSilentFallback`).
- Live-verified against a real `docker compose` stack (Postgres, MinIO):
  - Created a repo backed by a dedicated MinIO bucket, uploaded a real file, exported, deleted the repo, restored through the real UI (Chrome) — same `blobStoreId`, byte-identical content, blob physically back in the same S3 bucket.
  - Configured scheduled backup through the real UI (Chrome), including a full page reload confirming server-side persistence.
  - A real 1-minute cron schedule during verification drove 3 consecutive real runs; retention correctly kept only the 2 most recent (confirmed via `mc ls` against MinIO directly).
  - Confirmed a real Security → Audit Log entry for a successful scheduled run.


skensell201 (changes requested):
The per-asset storeFor fix and the scheduler wiring both read well, and the hard-error-vs-fallback distinction for the destination store is exactly right. One blocker and two smaller things, found while running the whole batch in one worktree.

## Blocker: blob GC will delete every scheduled backup

RunScheduled writes the archive as an ordinary key (`backups/nexspence-backup-<ts>.tar.gz`) into a regular blob store via `store.Put`. BlobGCService.compactStore (internal/service/gc_service.go:233) walks `store.ListEntries` and deletes any key that is not in the reference index — and that index is built purely from asset rows (`Assets.ListAllBlobRefs`). A backup archive never has an asset row, so as soon as it is older than `gc.min_age` the next Compact (manual or the GC cron) removes it. Retention of 7 daily backups can never be reached on an instance where GC runs; the three-runs-in-a-minute live check would not have hit this because nothing was past min_age yet.

The local store makes it even less obvious: keyPath shards the key, so the archive lands at `<base>/ba/ck/backups/nexspence-backup-….tar.gz` and ListEntries hands it back as a key GC considers an orphan.

Options, in the order I'd prefer:
1. Teach GC about the reserved prefix — skip `backups/` in compactStore (one `strings.HasPrefix` before the `referenced.has` check) and cover it with a test in gc_service_test that puts an unreferenced `backups/x.tar.gz` entry into the store, runs Compact past min_age, and asserts it survives. Document the reserved prefix on BlobStore/ListEntries so the next writer of a non-asset key knows GC ignores it.
2. Alternatively refuse a destination store that any repository uses and only accept a dedicated one — but that still leaves the operator one `Compact` click away from losing backups on that store, so I'd rather make GC prefix-aware.

## Migration number

034 was taken by #493 (blob_store_migrations FK), merged first, and #495 is taking 035 — goose panics on a duplicate version. Please renumber to 036 (037 if #450 lands before you rebase; I'll note which in the thread).

## Non-blocking

- applyRetention lists the *entire* store to find the `backups/` entries. On an S3/Azure store that also holds repository data that is a full bucket walk every night just to prune a handful of archives. The interface has no prefix listing today, so I'm fine with this for the first cut, but a `ListEntriesWithPrefix` (S3/Azure support it natively) would be the natural follow-up — worth a TODO.
- `store.Put` bypasses RegisterStoredBlob, so the destination's used_bytes/quota accounting never sees the archives. Also acceptable for now, but worth a line in the docs so a quota on that store is understood as 'artifacts only'.

skensell201:
Merge-order update: #495 took 035 and #450 took 036, so the backup_settings migration should become **037**.

jorgemillan:
Thanks for the review — all four points addressed in `2b36641`.

**Blocker (GC deleting scheduled backups):** confirmed and fixed. `BlobGCService.compact` now skips any key under the `backups/` prefix before evaluating it as an orphan (`internal/service/gc_service.go`), so scheduled backups are never candidates for collection regardless of `min_age`/schedule. Added `TestGC_ScheduledBackupSurvivesPastMinAge` as a regression test.

Went with the prefix-skip over restricting the destination to a dedicated store, since `CompactAll` iterates every blob store unconditionally today — there's no way to opt a store out of GC, so a "dedicated" destination would still get swept on the weekly run.

**Migration renumbering:** `034_backup_settings.sql` → `037_backup_settings.sql` (034/035/036 were taken by #493/#495/#450, all merged after this branch was cut). Applies cleanly on top of them.

**Quota accounting:** `RunScheduled` now calls `BlobStores.UpdateUsedBytes` on a successful write, and `applyRetention` does the same (negative delta) on each retention delete — same pattern the rest of the codebase already uses (`base.RegisterStoredBlob`, GC's own decrement).

**Retention's full-store scan:** added an optional `storage.PrefixListableStore` interface (same shape as `PresignableStore`/`AppendableBlobStore`), implemented for S3 and Azure using their native `Prefix` filter, and for Local by walking just the relevant shard directory instead of the whole tree. `applyRetention` uses it when the resolved store supports it, falling back to the existing full-list+filter otherwise.

**Verification:** `go build`/`go vet`/`go test ./...` green, plus `-tags integration` (real MinIO + Azurite via dockertest) for the two new prefix-listing implementations. Also ran the full feature live against a real `docker-compose` stack (Postgres + MinIO, GC and backup cron both on 1-minute schedules): multiple scheduled backups survived several real GC passes while a genuine orphan blob in the same store was correctly detected and removed in the same run; `usedBytes` tracked the live backup set correctly through several write+retention-delete cycles.

PR #491, opened 2026-09-17
fix(storage,frontend): suggest and verify a writable path for new local blob stores

n00bi14:
The "New Blob Store" form proposed a hardcoded `./data/blobs/<name>`. That path
is relative to the server's working directory and has nothing to do with
`storage.local.base_path`, so it is wrong in every deployment that configures a
different blob root. Under the Helm chart it is actively broken: the blob PVC is
mounted at `storage.local.mountPath` (`/blobs`), the working directory is `/app`,
and `readOnlyRootFilesystem: true` makes everything outside the mounts read-only
— the suggested path resolves to `/app/data/blobs/<name>`. The store was created
without complaint and the first artifact push failed, a long way from the setting
that caused it. The seeded `default` and `docker` stores escape this only because
`syncBlobStorePaths` rewrites them to `<base_path>/<name>` on every start; a store
created later never passes through it.

`GET /api/v1/system/info` now reports the effective storage configuration, and
the form proposes `<base_path>/<name>`, following the name field until the
operator edits the path themselves. The inline route closure moved into
`SystemHandler.Info` so it can normalize empty config values the way the storage
layer does at runtime and carry unit tests; the frontend treats the new block as
optional, so a UI served against an older server still renders.

Proposing the right path is not enough on its own, because the path can still be
typed or scripted. Creating a local store, and repointing an existing one, now
probes the path — create the directory, write a temporary file, remove it — and
refuses with 400 when that fails, naming EROFS or a permission problem
explicitly: the kernel's bare "read-only file system" tells an operator nothing
about which of their mounts is writable. The connection test had the same blind
spot from the other side, since `Exists()` is a read and a directory that exists
on a read-only volume answered "ok". `Update` probes only when the path actually
changes, so a quota edit does not start depending on the volume being mounted,
and an empty store type counts as local to match `newFromDescriptor`.

Deliberately not done: mounting the blob PVC a second time at `/app/data/blobs`.
That makes the wrong path work while leaving it wrong, and it makes a worse
mistake reachable — `PhysicalStoreIdentity` keys local stores as `local|<path>`
with no cleaning or mount resolution, so two logical stores addressing one
directory through two paths are separate namespaces to the GC, which is the
aliasing acd41e9 removed for Azure containers. The chart is unchanged and still
mounts the volume once.

`docs/deployment.md` and the Helm README now say that additional local stores
have to live under the configured blob root, which is the part that was only
discoverable by hitting the failure.

Tested with `go test ./...`, `npx tsc --noEmit` and `npm test`. One caveat worth
knowing: the negative-path tests build an unwritable directory with `chmod 0555`
and skip themselves when the process can write to it anyway, so on a CI runner
that executes the suite as root those four cases do not run and only
`classifyLocalPathErr` covers the error-classification branches directly.


skensell201 (approved):
Reviewed: the probe (MkdirAll + CreateTemp + Remove) is the right shape — it is exactly what the first Put would do, just moved to where the operator can still fix the path — and gating Update on an actual path change keeps quota edits independent of the volume being mounted. Naming EROFS explicitly is a real UX win under readOnlyRootFilesystem. DefaultLocalBasePath in storage (rather than config) avoids the import cycle and replaces three string literals. /api/v1/system/info stays admin-gated, which is where the create form lives. Verified as part of today's batch in one worktree (build/vet/lint/unit/integration/frontend all green). Merging.

PR #492, opened 2026-09-17
feat(helm): expose outbound.allowedInternalCidrs

fallen-up:
## Summary
- Add `config.outbound.allowedInternalCidrs` to the Helm chart.
- Emit `NEXSPENCE_OUTBOUND_ALLOWED_INTERNAL_CIDRS` (comma-separated) so the SSRF guard can reach an internal Nexus for migration/proxy/webhooks.
## Test plan
- [x] `python3 scripts/check-helm-env.py`
- [x] `helm template` with a CIDR set and grep the ConfigMap for `NEXSPENCE_OUTBOUND_ALLOWED_INTERNAL_CIDRS`
- [x] migration preview against an internal Nexus IP succeeds when the CIDR is set


skensell201 (approved):
Reviewed: the Helm value maps to the existing config key outbound.allowed_internal_cidrs (internal/config/config.go:89), the env name NEXSPENCE_OUTBOUND_ALLOWED_INTERNAL_CIDRS follows viper's key derivation, and scripts/check-helm-env.py gained a render case so the guard covers it. Verified together with the rest of today's batch in one worktree over main (build/vet/lint/unit/integration/helm guard all green). Merging.

PR #493, opened 2026-09-17
fix(storage,db,api): let a blob store be deleted once nothing references it

n00bi14:
## Problem

Deleting a blob store failed with a raw driver error, even when no repository used it any more:

```
ERROR: update or delete on table "blob_stores" violates foreign key constraint
"blob_store_migrations_target_store_id_fkey" on table "blob_store_migrations" (SQLSTATE 23503)
```

Two independent defects met here.

**Migration history pinned the store forever.** Migration 015 created `blob_store_migrations` with foreign keys to `blob_stores` and no `ON DELETE` clause, so Postgres' default `NO ACTION` applied. Any finished migration row — including one whose repository was deleted long ago, since `repository_name` is `TEXT` with no FK by design — blocked `DELETE` on the target store permanently. There was no way out: the repo has no `Delete` for history rows, and `DELETE /api/v1/repositories/:name/blob-store-migration` is Cancel, which only sets a status. An operator's only option was manual SQL.

**Every blocker surfaced as an HTTP 500 with SQL internals.** `BlobStoreHandler.Delete` checked group membership and nothing else, so referencing repositories, leftover assets and history rows all fell through to the driver error, which the admin UI printed verbatim (screenshot above).

## Changes

**Schema** — new migration `034_blob_store_migration_fk_on_delete.sql`: both FKs become `ON DELETE SET NULL` and `target_store_id` becomes nullable. The audit row survives the store; only the pointer is cleared. `CASCADE` would have been wrong — deleting a store must not erase the record of how artifacts moved through it. `Down` is lossy by necessity (rows already `NULL` cannot satisfy the restored `NOT NULL`) and says so.

**Persistence** — `blobStoreMigrationCols` now `COALESCE`s `target_store_id`, which would otherwise break `scanMigration` on the first `NULL`. `Create` still refuses a migration without a target: an empty target is a historical state, never a new one.

**Service** — `ResumeAll` and `runMigration` fail a job whose destination store is gone, with a message saying so, instead of looking up an empty store id.

**Error translation** — `ErrInUse` / `InUseError` in `internal/repository`, a `foreignKeyViolation` helper alongside the existing `uniqueViolation` in `postgres/errors.go`, and `blobStoreRepo.Delete` translating SQLSTATE 23503 instead of leaking it.

**API** — `Delete` gains pre-checks for repositories (`ListByBlobStoreID`) and assets (`CountByBlobStoreID`, needed because a group member can hold artifacts while `repository.blob_store_id` points at the group) and answers **409** naming the blocker. The translated FK error is the backstop for anything a pre-check misses, so no delete path returns a 500 carrying a constraint name any more. `docs/api-spec.yaml` documents the 409.

**Docs** — `docs/blob-store-groups.md` gets a "Deleting a blob store" section; `docs/schema.sql` finally records `blob_store_migrations`, which it had been missing since 015.

## Upgrade notes

Existing installations need nothing beyond `goose up`. Replacing the constraint applies to rows already in the table, so a store that is undeletable today becomes deletable as soon as 034 has run — no backfill, no manual SQL.

## Behaviour deliberately left alone

- **History is never pruned.** Rows for deleted repositories stay. They are the audit trail — which repository moved, when, how far it got, and why it failed — and after 034 they block nothing. Only rolling 034 back deletes any.
- **Deleting a store still does not delete blobs.** It removes the catalog row; objects on the backing disk, bucket or container are left in place. This is pre-existing behaviour, now written down: run Compact first if you want unreferenced objects gone, because once the store row is deleted its path or container is no longer known to the application.


skensell201 (approved):
Reviewed: SET NULL on both FKs is the correct trade-off (history survives, pointer clears), the COALESCE in blobStoreMigrationCols plus the Create guard keep an empty target strictly historical, and ResumeAll/runMigration failing a job whose target is gone closes the hole SET NULL would otherwise open. The two-layer approach in Delete — pre-checks for repositories and assets producing a named 409, and the SQLSTATE 23503 translation as a backstop — means no delete path can leak a constraint name any more.

Migration numbering: three PRs opened this week all took 034 (this one, #490, #495) — goose panics on a duplicate version when they meet. This PR merges first and keeps 034; the others will be renumbered. Verified in one worktree with the whole batch (postgres + e2e integration green after renumbering). Merging.

Issue #342, opened 2026-08-27
Nexus Migration: 10 correctness/data-loss bugs found migrating from a real Nexus instance


jorgemillan:
We ran the Nexus Migration feature end-to-end against a real, populated Nexus OSS instance (47 repositories, 1000+ security roles, 73 non-built-in privileges, 8 content selectors, 14 local users) rather than a synthetic fixture, and found 10 distinct issues — several of them silent, i.e. the migration reports success while dropping data or permissions. All 10 are fixed and verified below (unit tests + live runs, several re-confirmed against the real instance after a full stack reset). Filing together since they share one feature area and several interact with each other.

---

## 1. A fatal error in any later migration stage discards the success of earlier stages, with no way to skip just the broken stage

**Labels:** bug, high, migration

### Summary

`runStages` walks `MigrateRepos→MigrateBlobs→MigratePrivileges→MigrateRoles→MigrateUsers→MigrateRoutingRules` as a strict sequence. Each stage tolerates per-item failures internally, but returns a hard error when it "cannot proceed at all" — and `runStages` propagated that hard error straight up, aborting every stage after it.

### Where

- `internal/service/nexus_migration_service.go` — `runStages`, and `run`'s error-to-status mapping
- `internal/api/handlers/migration.go` — no endpoint to update an existing job's scope flags without deleting and recreating it

### Reproduction

Reproduced organically against the real instance: its `/security/users` endpoint has a genuine data bug (`TooManyResultsException: Expected one result... found: 2`, i.e. a duplicate LDAP record), returning `500`. Before the fix: `MigrateRoutingRules` never even ran, despite depending on nothing that failed. After the fix: routing rules still run, `lastError` is prefixed per stage (`"users: list users on the source Nexus: ..."`), and the job still correctly ends in `error` — but with every independent stage having completed.

### Impact

If e.g. `migrateUsers` fails because of a source-side data bug, that one error from the *fifth* of six stages is the only thing reflected in the job's final status, even though repos/blobs/privileges/roles may have already migrated successfully — and routing rules never get a chance to run at all, despite having no dependency on users succeeding.

### Fix

[...]

PR #495, opened 2026-09-17
fix(migration): artifacts-only, per-repo scope, no silent empty success

### LINKED ISSUE #494: Nexus migration cannot do a real cutover: silent empty success, no artifacts-into-existing-repos, no per-repository scope
## Summary

The live Nexus -> Nexspence migration job has three related holes that make a real cutover either a silent no-op or an all-or-nothing full-instance copy.

1. **A job can finish `done` while copying nothing.** The Artifacts stage is nested inside the Repositories stage. Unchecking Repositories (and leaving Artifacts on) still returns a successful job with `assetsDone: 0`. The operator sees OK; under the hood the blobs walk never ran. The same empty-success shape shows up when the source is unreachable (SSRF guard, DNS, credentials): the UI lets you Start without Test connection, the runner cannot list anything, and the card looks like a finished empty run rather than a failed one.

2. **Artifacts cannot be copied into hosted repositories that already exist.** Destinations are often created first — by hand, by Terraform, by a previous job — and only the bits need to come across. Today that is impossible: Artifacts does not run unless Repositories is also on, and Repositories is a create-if-absent of the *definitions*, not a blob transfer into what is already here.

3. **There is no per-repository scope.** A job always walks every repository on the source. There is no way to migrate one hosted repo, a handful, or to catch up a repo that was already created here. Resume already skips assets that exist, so a named allowlist would also be the catch-up path; it just is not exposed.

## Where

- `internal/service/nexus_migration_service.go` — `runStages` (Artifacts gated on `MigrateRepos`), `migrateRepositories`, `migrateAssets`
- `internal/api/handlers/migration.go` — `POST /api/v1/migration/jobs` has no repository allowlist; `POST /api/v1/migration/preview` is optional
- Admin → Migration wizard (`frontend/src/pages/AdminPage.tsx`) — Start does not require Test connection; no repository picker
- `internal/domain/migration.go` / `migration_jobs` — no column for the selected source names

## Impact

A cutover that should have been "copy the artifacts in these hosted repos" either:

- reports success and transfers zero bytes (Artifacts without Repositories), or
- forces a full-instance sync of every hosted/proxy/group plus the security model, which is what operators then have to pause, clean up, and redo.

An unreachable source is easy to mistake for "there was nothing to copy."

## Suggested fix

- Run Artifacts independently of Repositories. When Repositories is off, copy into matching **hosted** destinations that already exist here. A missing dest, a dest that is not hosted, or a format mismatch is a per-item error and a skip — not a write into the wrong repo, and not a silent zero.
- Add `scope.repositories` (source names). Empty/omitted keeps today's "every repository" contract for API clients. A group in the list should include its members: a group cannot be created empty, and artifacts live on the hosted members.
- In the Admin wizard, Repositories/Artifacts cannot Start until Test connection succeeded, so a blocked or down source cannot look like an empty success. Show a picker of the previewed names.
- Persist the allowlist on the job row so resume/restart still copies only what was asked for. Existing assets stay skipped, which is the catch-up.

fallen-up:
## Summary
- Artifacts no longer depend on Repositories. A blobs-only job copies into destinations that already exist here; a missing dest or a type/format mismatch is an error and a skip, not a silent empty success.
- `scope.repositories` limits the job. Empty or omitted still means every repository. A selected group also migrates its members (a group cannot be created empty).
- Artifacts copy from **hosted repositories and proxy caches**. Groups have none of their own. A proxy cache is copied rather than re-fetched, so a dead or archived upstream does not leave the new proxy empty. Destination type must match the source (hosted↔hosted, proxy↔proxy).
- Hosted PyPI files under Nexus `packages/<name>/<version>/<file>` are rewritten into Nexspence `/packages/<name>/<file>` and copied even when the Components API omits them. `/simple/<pkg>/` is rebuilt from stored wheels (`ListByComponentID`), not from the first 100 assets of the whole repo.
- Generated catalogs are not transfer errors on hosted repos (Nexspence rebuilds them): Helm `index.yaml`, Maven `maven-metadata.xml` and checksums, npm packuments, PyPI `/simple/` pages. On a **proxy** those same files are part of the cache and **are** copied.
- Admin wizard: Test connection is mandatory before Repositories/Artifacts; Next is disabled until then. After Test, nothing is selected by default. Artifacts-only greys groups only; proxies stay selectable. Job cards show the picked name(s), hide the Repositories counter when that stage is off, and history expands `lastError` under that row.
Fixes #494

## Test plan
- [x] Artifacts on, Repositories off, dest missing → job `done`, `errorCount > 0`, dest not created, 0 assets
- [x] Same, dest already hosted → assets copy; dest definition unchanged
- [x] Dest exists as proxy / other format → skip, no write
- [x] `scope.repositories: ["npm-local"]` → only that repo (and its artifacts); others untouched
- [x] Group in the list → members created/copied too
- [x] API omits `repositories` → full instance (old contract)
- [x] Admin: Start without Test connection blocked; 502 from preview stays an error, not “0 repos”
- [x] Resume of a named job still skips assets that already exist (catch-up)
EOF


skensell201 (approved):
Reviewed the whole change (service, handler, migration, PyPI handler, wizard/picker) and verified it in one worktree with the rest of today's batch — build/vet/lint/unit/integration/frontend all green.

What convinced me:
- The three outcomes are now explicit: Artifacts runs on its own, acceptBlobDest refuses a missing / wrong-type / wrong-format destination as a per-item error, and reportMissingAllowlist turns an allowlist typo into a visible lastError instead of a silent zero.
- expandRepoAllowSet resolves group members from SOURCE data (nested groups included), consistent with how ensureRepository already flattens groups (#342).
- Copying a proxy cache whole (appendUnplannedAssets, catalogs included) vs. skipping generated catalogs on hosted is the right split — on a proxy those files *are* the cache.
- The PyPI servePackageIndex change from Assets.List(…, 100, 0) to ListByComponentID fixes a real first-100-assets bug on its own.
- Empty/omitted scope.repositories keeps the old contract; the TEXT[] DEFAULT '{}' column keeps existing jobs meaning 'everything'.

Two housekeeping things I pushed to your branch rather than round-tripping (maintainerCanModify): merged main (the AdminPage import line collided with the batch) and renumbered the migration 034 → 035, since #493 took 034 and merged first — goose panics on a duplicate version. Merging once CI is green.

PR #496, opened 2026-09-17
fix(browse): scroll the component list instead of clipping it

fallen-up:
## Summary
- Browse clipped a page of components: `.holo-card` overflow:hidden let the table shrink with no scrollbar.
- The list scrolls; pager stays put. Same for docker/raw trees.
## Test plan
- [x] Open Browse on a hosted repo with a full page of components (25).
- [x] All rows reachable without changing browser zoom.
- [x] Prev/Next still visible.
- [x] Docker/raw trees still scroll.


fallen-up:
image

skensell201 (approved):
Reviewed: the root cause (a flex item with overflow:hidden shrinking below its content because of the min-height:auto default) is right, and the fix threads flex:1/minHeight:0/overflow:auto through page → table/tree/detail panel while pinning header, toolbar, selection bar and pager with flexShrink:0. The new BrowsePage test locks the style contract; frontend typecheck/test/build green on the combined batch tree. Merging.

PR #497, opened 2026-09-18
feat(repositories): filter by type

fallen-up:
## Summary
- Repositories toolbar keeps one row: name search, All/Hosted/Proxy/Group, All formats.
- Type, name, and format filters combine (AND).
## Test plan
- [x] All shows every repo; Hosted/Proxy/Group each hide the other types.
- [x] Type + name + format stack.
- [x] Search still shrinks; the type control and format dropdown stay on the same row.


fallen-up:
image
---->
image

I considered using another dropdown for the type filter, but with only four options (All, Hosted, Proxy, Group) I think the extra click would make the UX worse. Keeping the options visible makes switching faster, and there is still plenty of horizontal space left for the search field.

skensell201 (approved):
Reviewed: purely client-side filter over the already-loaded repository list, radiogroup/aria-checked semantics on the pills, a proper empty state when nothing matches, and the existing 'type pills' assertions were correctly relaxed to getAllByText since the labels now appear twice. Frontend typecheck/test/build green on the combined batch tree. Merging.

PR #499, opened 2026-09-18
fix(api): require nx-admin on eight read-only admin-area endpoints

### LINKED ISSUE #498: Eight read-only admin endpoints missing the `nx-admin` role check
## Summary

`internal/api/router.go` registers most sensitive GET routes under the `admin` group (`authMW` + `AdminRequired()`), but eight were left under `authed` (any authenticated user, any role) instead — a role check present on every sibling write route in the same handler family was simply missing on the read side.

A black-box + gray-box pentest against a real deployment (2026-09-18) found that a plain `nx-developer` (or `nx-deployer`) session could read: JSON dashboard metrics, the role catalog, the privilege catalog, the privilege→role map, content selectors, and replication rules — while `security/users`, `audit`, `blobstores` write ops, `migration/jobs` and `backup/export` correctly required `nx-admin` for the exact same caller.

Two endpoints the pentest also flagged turned out **not** to be bugs on closer inspection, and are intentionally excluded here:
- `GET /service/rest/v1/blobstores` — `domain.RedactedBlobStore` already strips every credential (`secret_key`, `account_key`, `connection_string`, `sas_token`) before the config is serialized, and `RepositoriesPage.tsx` genuinely needs blob-store names for any authenticated user (repo list rendering).
- `GET /api/v1/system/services` / `GET /api/v1/security/scanner` — already `admin`-gated; the internal hostnames/paths they return are the intended content of an admin health-check page, and credentials are already stripped (`dsnDetail` only returns `host/db`, never the DSN's userinfo).

## Where

`internal/api/router.go`:
- `authed.GET("/api/v1/metrics"...)`, `/api/v1/metrics/history`, `/api/v1/metrics/repos`
- `authed.GET("/service/rest/v1/security/roles", roleH.List)`
- `authed.GET("/service/rest/v1/security/privileges"...)`, `/:id`, `/service/rest/v1/security/roles/:id/privileges`, `/api/v1/security/privilege-role-map`
- `authed.GET("/service/rest/v1/security/content-selectors"...)`, `/:id`
- `authed.GET("/api/v1/replication/rules"...)`, `/:id/history`

## Impact

Low-to-medium: no endpoint returns raw credentials (`ReplicationRule.TargetPasswordEnc` already carries `json:"-"`), but it's a real least-privilege violation — any authenticated user, including a deploy-only service account, could enumerate the role/privilege catalog and replication targets, and see live server metrics meant for the admin monitoring dashboard.

## Suggested fix

Move the eight `authed.GET(...)` registrations into the `admin` group, next to their write-side counterparts. Confirmed via `frontend/src/api/client.ts` + a page-level search that none of these eight are called from any non-admin-only page (`SecurityPage`, `UsersPage`, `AdminPage`/`MonitoringPage` are the only callers) — moving them breaks nothing in the UI.

**Verified 2026-09-18** end-to-end against a real docker compose stack (Postgres, MinIO, Jaeger/OTel) with real `nx-developer`/`nx-deployer` accounts: 403 on all eight before the fix would have been 200, 200 for `nx-admin`, untouched sibling routes unaffected. Playwright-checked the admin UI (Security/Monitoring/Replication tabs) for regressions — none found.

jorgemillan:
### Body

Closes #498.

## Summary

`internal/api/router.go` registers most sensitive GET routes under the `admin` group (`authMW` + `AdminRequired()`), but eight were left under `authed` (any authenticated user, any role) — the same role check every sibling write route already has was simply missing on the read side. A pentest against a real deployment (2026-09-18) found any `nx-developer`/`nx-deployer` session could read:

- JSON dashboard metrics (`/api/v1/metrics`, `/metrics/history`, `/metrics/repos`)
- the role catalog (`/service/rest/v1/security/roles`)
- the privilege catalog + privilege→role map (`/security/privileges`, `/security/roles/:id/privileges`, `/api/v1/security/privilege-role-map`)
- content selectors (`/service/rest/v1/security/content-selectors`)
- replication rules (`/api/v1/replication/rules`, `/:id/history`)

while `security/users`, `audit`, `blobstores` writes, `migration/jobs` and `backup/export` correctly required `nx-admin` for the exact same caller.

**Deliberately not touched**, on closer inspection during the same pentest:
- `GET /service/rest/v1/blobstores` — `domain.RedactedBlobStore` already strips every credential before serializing, and `RepositoriesPage.tsx` genuinely needs the list for any authenticated user (repo-row blob-store names). Gating it would have broken that page.
- `GET /api/v1/system/services`, `GET /api/v1/security/scanner` — already `admin`-gated; the internal detail they return is the intended content of an admin health-check page, and `dsnDetail()` never returns the DSN's userinfo.

## Change

Move the eight `authed.GET(...)` registrations into the `admin` group, next to their write-side counterparts — no new code, just re-registering existing handlers under the group that already exists for this exact purpose.

## Verification

Confirmed via `frontend/src/api/client.ts` + a page-level search that none of the eight are called from any non-admin-only page (`SecurityPage`, `UsersPage`, `AdminPage`/`MonitoringPage` are the only callers) — moving them breaks nothing in the UI.

Verified end-to-end against a real `docker compose` stack (`--profile monitoring --profile tracing`: real Postgres, MinIO, Jaeger/OTel collector receiving live traces from the running app):
- Created real `nx-developer` and `nx-deployer` accounts via the live API and hit all eight endpoints: **403** for both (was 200 before this fix), **200** for `nx-admin`.
- Confirmed the untouched sibling routes (`blobstores`, `cleanup-policies`, `repositories`, `me/privileges`) still answer **200** for non-admin — no regression.
- Playwright (Chromium): admin's Security/Monitoring/Replication tabs still render with no console errors; a `nx-developer` session hitting `/security` or `/repositories` directly degrades gracefully (no error boundary), and `/repositories` still renders correctly via its own intentionally-unrestricted `listBlobStores()` call.
- `go build ./...` and `go test ./...` green.

Test users deleted and the compose stack torn down after verification; no state left behind.


skensell201 (approved):
Verified the caller claim independently: the only frontend users of these eight are AdminPage/MonitoringPage/UsersPage/SecurityPage, and every one of those pages already calls an admin-gated endpoint on load, so a non-admin session gets nothing new broken. Verified as part of today's batch (all open PRs merged into one worktree: build/vet/lint/unit/integration green). One thing I'd still like — not blocking — is a regression test in internal/integration that logs in as nx-developer and asserts 403 on these paths; I'll add it as a follow-up so the router grouping cannot drift back silently. Merging.

PR #517, opened 2026-09-19
fix(docker): allow dots in subdomain alias targets

fallen-up:
## Summary
- Hostname aliases whose target is a dotted repository name (`quay.io`, `ghcr.io`, `registry-1.docker.io`) were dropped at rewriter construction: `isRepoNameLabel` only allowed a DNS label, so `/v2/strimzi/…` on the alias host was dispatched as repository `strimzi` and anonymous pull returned 401.
- Alias targets now use `isRepositoryName` (dots allowed; `/` and `..` still rejected). #478 already kept dotted *host* keys in the YAML map; this is the matching check on the *value*.
- Path-style pulls on `baseDomain` are unchanged. Do not shorten `baseDomain` to a parent of the UI host: `nexspence.example.com` would then be rewritten as subdomain `nexspence`.

## Test plan
- [x] `go test ./internal/api/ -run TestSubdomainRewriter_`
- [x] Alias host `docker-quay-proxy…/strimzi/kafka:<tag>` rewrites to `/v2/quay.io/strimzi/…` and pulls
- [x] Path-style `<baseDomain>/quay.io/strimzi/kafka:<tag>` still pulls


skensell201 (changes requested):
Thanks — the diagnosis is right and the test covers the reported case. One gap before merging:

`isRepositoryName` is a hand-rolled grammar that is stricter than the one `RepositoryService.Create` enforces for docker/oci names (`dockerPathComponent` in `internal/service/repository_service.go`: `^[a-z0-9]+(?:(?:[._]|__|[-]+)[a-z0-9]+)*$`). It rejects `_`, `__` and `--`, all of which Create accepts (`team.registry_2`, `a--b` are in the service tests). So an alias pointing at such a repository is still silently dropped at rewriter construction — the same failure mode this PR fixes for dots, just for a different character.

Suggestion: use the same regex here (it already excludes `/`, `..`, leading/trailing separators and empty components, so the manual `ContainsAny`/`prevSep` checks become unnecessary), and add `team.registry_2` to the alias-target test. Keeping one grammar for "what is a docker repository name" also means the two can't drift apart again.

fallen-up:
Done — the grammar is shared now. `dockerPathComponent` moved to `internal/domain` as `IsDockerPathComponent`, and both `RepositoryService.Create` and `NewSubdomainRewriter` call it, so `team.registry_2` and `a--b` are usable alias targets and the two checks cannot drift apart. The manual `ContainsAny`/`prevSep` logic is gone, as you suggested — the regex already covers slashes, `..`, empty components and leading/trailing separators.

Tests: `team.registry_2` and `a--b` added on the alias side, plus a table test for the grammar itself in `internal/domain`.

One thing I left alone: the subdomain path still uses the DNS-label check (`isRepoNameLabel`). That value comes from the Host header, where `_` and `.` cannot appear in a single label anyway, so widening it there would only add reach for hosts DNS will not serve.

Rebased onto current main.

